Issue #10, opened 2026-09-05
Probe: identity check (throwaway)


nateprich:
**This is a throwaway diagnostic, not real work.** It will be closed and its branch deleted once read. Do not treat it as part of Command Center v0.

## What to do

Three things, in this order:

1. **Assign this issue to yourself**, if you are able to.
2. **Leave a comment** on this issue saying which tool and account you are acting as.
3. **Open a pull request** that adds a file `probe.txt` at the repo root containing the single word `ok`. Base it on `main`. Do not merge it.

Nothing else. Do not modify any other file.

## Why

Command Center needs a single-in-motion lock so that only one agent run is ever working at a time. The intended mechanism is "an open issue assigned to the agent", which only works if the agent's GitHub writes are distinguishable from Nate's own.

Across 709 issue events and 197 pull requests in these repositories, every single write shows `actor=nateprich` with no originating app recorded. That may be because a connected app acts as the account holder, or simply because this agent has never run here before. Those two explanations are indistinguishable from the existing history, so this probe settles it by generating one write of known origin.

The measurement is `actor.login` and `performed_via_github_app` on the resulting events — so **the act of doing the three steps is the entire point**; the content is irrelevant.

nateprich:
Acting via Codex desktop using the GitHub CLI, authenticated as @nateprich.

nateprich:
Answered. Codex acts as @nateprich via the gh CLI — `app=none` on every write, no bot identity. Kept closed rather than deleted so the evidence trail survives; the finding is written up in LEARNINGS.md.

PR #11, opened 2026-09-05
Add identity probe file

nateprich:
Probe requested by #10. Adds only `probe.txt` containing `ok`.


nateprich:
Probe complete. Codex acts as @nateprich via the gh CLI; assignment cannot serve as the single-in-motion lock. Evidence recorded in LEARNINGS.md. Closing unmerged — probe.txt was never real work.

PR #36, opened 2026-09-06
Add explicit reasoned parking

### LINKED ISSUE #19: funnel park <ref> --reason: refuse without a reason, park with one
Part of #16. Read [`plan.md`](plan.md) first — the funnel table and its rejected alternatives are the spec, not just the issue body.

## What to build

A `park` subcommand in `funnel.py`, alongside `claim` / `release` / `reject`:

```
funnel park <ref> --reason "<why>"
```

`<ref>` takes the same forms as `claim` — issue number, `owner/repo#number`, or URL. `find()` already handles all three.

In order:

1. **Refuse without a reason.** Missing `--reason`, an empty string, and a whitespace-only string all refuse with a non-zero exit and a message saying what is missing. **Nothing is written on a refusal** — no Status change, no close, no comment. A half-applied park is worse than no park.
2. **Set `Status` to `Parked`** on the Project item, using `SET_FIELD` / `_option_id`, the way `cmd_reject` sets `Building` / `Broken`.
3. **Close the issue as not planned** — `state_reason: not_planned`. `plan.md` requires Done and Parked to stay distinguishable and `state_reason` is the mechanism; closing as completed makes a parked idea read as shipped and destroys the finished-vs-abandoned ratio the plan wants to measure.
4. **Post the reason as a comment**, so it is readable without opening the Project.

Post the comment with a fixed machine-readable prefix, as a module constant:

```python
PARK_COMMENT_PREFIX = "**Parked:** "
```

The sibling ticket reads the reason back out of that comment, so the marker is a shared contract, not decoration. Introduce the constant here if it does not exist yet.

## How to tell it worked

Tests in `tests/`, in the style of `test_ordering.py` — fixtures and fakes, no network:

- `--reason` absent → non-zero exit, and no GitHub call attempted
- `--reason ""` → same
- `--reason "   "` → same
- a real reason → `Status` written as `Parked`, the close called with `not_planned`, and a comment posted whose body contains the reason text behind `PARK_COMMENT_PREFIX`

## Constraints, from the plan

Violating one of these fails review even if the code works.

- **Do not add a label.** The set is closed at two (`needs-shaping`, `blocked`) and that is deliberate.
- This becomes the **second** place `Status` is written by code rather than by Nate. `cmd_reject`'s docstring currently claims to be the only one — update that sentence so it stays true.
- **Do not touch `maintenance_load`.** Its `NOT_PLANNED` exclusion (`funnel.py:388`) is deliberate: parking is a decision, not work done.
- **No auto-parking.** `plan.md` rejects age-based or bulk parking outright: parking is itself a clearing action, the bottom-up ordering already forces the decision, and auto-park would write the reason line for you — losing the only part with durable value. Do not add a `--all`, a staleness threshold, or a default reason.

## Scope

Stops at the command. Surfacing parked items in `funnel brief` is the sibling ticket. Neither blocks the other — that one can be tested from fixtures alone.

## Two places that go stale if you skip them

Both are a line each, and both are wrong the moment this lands:

- `README.md` has a table of `funnel` commands. Add the `park` row.
- `skills/funnel/SKILL.md` already tells the reader to *"name parking as a move"* on a long-waiting item, with no command to point at. Point it at `funnel park`.

nateprich:
## Summary

- add `funnel park <ref> --reason <why>`
- reject missing, empty, and whitespace-only reasons before any GitHub read or write
- set Project Status to Parked, close as not planned, then post the reason with the shared `PARK_COMMENT_PREFIX`
- document the command in `README.md` and the funnel skill

## Deliberately not changed

- no labels, auto-parking, bulk mode, default reason, or staleness threshold
- `maintenance_load` and `funnel brief` are untouched

## Verification

- `python3 -m pytest tests/ -q` (175 passed locally; GitHub Actions pytest also passed)
- `git diff --check`
- command help verified locally

The original implementation created `ticket/19` from `main`. A subsequent scheduled run found the remote branch and this open PR, deliberately continued it rather than resetting it, reviewed the complete diff against the issue and `plan.md`, and found no additional changes necessary. No known departures from `plan.md` and no unresolved uncertainties.

Closes #19


nateprich (commented):
**Changes requested — not merged.** One item from the ticket's own constraint list is not satisfied.

Everything else checks out, so this is a one-line fix:

- behaviour matches the ticket step for step — refuse before any GitHub read, `Status → Parked` via `SET_FIELD`/`_option_id`, close with `not planned`, then the comment behind `PARK_COMMENT_PREFIX`
- the refusal really is pre-GitHub: `--reason` is validated by `_parking_reason` during `parse_args`, and `main()` does not reach `load_items()` until after that
- all four required test cases are present, and `python3 -m pytest tests/ -q` gives 175 passed locally (CI pytest green too)
- constraints honoured: no label, `maintenance_load` untouched, no `--all`/threshold/default reason
- the only files touched outside the ticket are `README.md` and `skills/funnel/SKILL.md`, both of which the ticket asks for by name; the `SKILL.md` line uses the sanctioned `/Users/nateprich/.claude/command-center` spelling

## What does not match

The ticket says of `cmd_reject`'s docstring: *"update that sentence so it stays true."* The replacement at `funnel.py:885` is not true.

```python
This is one of two commands that write `Status` by code rather than by Nate;
it remains the single case that also writes `Class`.
```

There are four commands that write `Status` by code, not two:

- `cmd_capture` → `Ideas` (`funnel.py:990`)
- `cmd_shaped` → `Shaped` (`funnel.py:1030`)
- `cmd_reject` → `Building` (`funnel.py:942`)
- `cmd_park` → `Parked` (new)

`cmd_answer` is the one that is Nate answering a gate.

Two further points, both of which argue for the same fix:

1. **The original sentence was about `Class`, not `Status`, and it was already true.** It read *"This is also the single case in the design where `Class` is written by code rather than by Nate."* Parking does not write `Class`, so nothing about this ticket made that sentence stale. The ticket's premise — that the docstring "currently claims to be the only one" in a sense parking invalidates — was mistaken. The second clause of your replacement preserves the true `Class` claim; it is the new `Status` clause that introduces the error.

2. **`plan.md` does not treat parking as code writing `Status` in Nate's place.** Line 461: *"only `Done` and `Parked` are now written by Nate answering a gate"*, and the parking prompt rides on the accept gate. `funnel park` is the command that records his decision, not a substitute for it. Describing it as written "by code rather than by Nate" contradicts that.

## The fix

Restore the original `Class` sentence in `cmd_reject`'s docstring verbatim. It was accurate before this change and is still accurate after it. Do not substitute a corrected count — a four-item tally in a docstring goes stale the next time a command writes `Status`, and this repo's standing rule is that a confidently wrong doc is worse than no doc.

If you would rather say something about parking, put it in `cmd_park`'s own docstring, where it can describe what parking is without making a claim about the rest of the module.

Nothing else needs to change. Re-push that one edit and this is mergeable.

nateprich:
Nate reviewed the finding and directed the merge as-is: the behaviour is correct and fully tested, and he is accepting the one inaccurate sentence in `cmd_reject`'s docstring rather than spending a round trip on it. Merging on his instruction, not on my review — recording that here so the override is visible rather than looking like the review was withdrawn.

PR #40, opened 2026-09-07
Surface parked items in the funnel brief

### LINKED ISSUE #20: Surface parked items and their reasons in funnel brief
Part of #16. Read [`plan.md`](plan.md) first.

## What to build

A written reason is only worth requiring if it is findable later — it is *"the artifact that makes re-encountering an idea in four months a 30-second decision rather than a re-derivation."* Today `funnel brief` does not mention parked items at all, so the reason is written and then buried.

Add a `parked` list to `cmd_brief`'s JSON: items whose `Status` is `Parked`, most recently parked first, each carrying ref, title, url, when it was parked, and **the reason**.

- "When it was parked" is already available as `status_since` — the `PROJECT_V2_ITEM_STATUS_CHANGED_EVENT` timeline the item query already fetches. No new field on the main query.
- The reason comes from the park comment, identified by the `PARK_COMMENT_PREFIX` marker the `funnel park` command writes. **Fetch comments only for items already known to be `Parked`** — the common path must not pay for a comment fetch on every item in the Project. If no marked comment is found, report the item with a null reason rather than omitting it; a parked item with a lost reason is exactly the thing worth seeing.

Then pin down, with tests, the two exclusions that keep parking a **clearing** action rather than a second queue:

- a parked item does **not** appear in `counts_by_gate`. It is closed, so the existing `state == "OPEN"` filter in `cmd_brief` should already drop it — the test exists so a later change cannot quietly reverse it.
- a parked item does **not** appear in `items` and is **not** counted in `total_needing_nate`. `gate_question` returns `None` for a closed item, so parking stops something waiting on Nate the moment it lands. Same reasoning: pin it.

## How to tell it worked

Add a parked item to `tests/fixtures/project_items.json` — `state: CLOSED`, `stateReason: NOT_PLANNED`, `Status: Parked`, with a park comment — and test that:

- it appears in `brief["parked"]` with its reason text
- it is absent from `counts_by_gate`
- it is absent from `items`, and `total_needing_nate` does not count it
- `maintenance_load` is unchanged by its presence

## Constraints, from the plan

- **Do not change the `maintenance_load` `NOT_PLANNED` exclusion** (`funnel.py:388`). Parking is a decision, not work done, and that line says so on purpose.
- **Do not add a label.**
- **Do not set `Status` or `Class` on anything.** This ticket only reads.

## Ordering

Does not depend on the `funnel park` command ticket. A parked item is one whose `Status` is `Parked` and which is closed as not planned, however it got that way — including by hand, which is how parking works today. The fixture supplies the comment, so this is testable with or without the command. Either ticket can be worked first; whichever lands first introduces `PARK_COMMENT_PREFIX`.

nateprich:
Closes #20

## What changed
- add a `parked` section to `funnel brief` JSON, ordered by most recent Parked transition
- read only the marked `**Parked:** ` comment for already-parked items, returning `null` when the reason is unavailable
- add fixture and regression coverage for ordering, lookup scope, queue/count exclusions, and maintenance metrics

## Deliberately not changed
- `counts_by_gate`, `items`, `total_needing_nate`, and `maintenance_load` semantics
- `funnel park`, labels, Project fields, or issue status/class values

## Verification
- `python3 -m pytest -q tests/test_brief.py tests/test_ordering.py tests/test_park.py` (69 passed)


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "47aaf381fa7186bb7e885cd443bdf34d44c0414f",
  "reviewed_at": "2026-09-07T04:01:55.011443+00:00",
  "verdict": "approved"
}
```

PR #94, opened 2026-09-07
Order Broken items within decision gates

### LINKED ISSUE #44: Make `Broken` sort first within gates; take `needs-shaping` out of ordering
Part of #32. Read `plan.md` ("Two queues, two orderings") first, then #32 — its Rejected list is load-bearing: a diff that reintroduces a rejected option fails review even if it works.

One concern: the sort keys. Two functions, both in `funnel.py`, tests with each.

### `awaiting_decision()` (`funnel.py:232`) — Nate's queue

Sort key becomes `(depth, not is_blocked, 0 if Broken else 1, since, repo, number)`. `Broken` enters **after** the bottom-up keys, so it breaks ties within a gate and never outranks a deeper gate — that ordering is Nate's decision recorded on #32, not a convenience.

`effective_class(item, by_ref)` (`funnel.py:191`) already exists but takes a `by_ref` map that `awaiting_decision()` does not currently build. Building that map (once, over `items`) is the one non-trivial part of this change — tickets inherit their Class from the parent, so a ticket in the queue resolves through its parent's Class.

### `ideas()` (`funnel.py:253`) — Ideas list

Drop the `needs-shaping` key; sort by `(0 if Broken else 1, status_since, repo, number)`. Everything in Ideas sits at one stage, so `Broken` first outright is safe here — this is the case that bit (#15 demoted for a missing label). The label itself stays (Nate keeps it for browsing; `AGENTS.md` closes the label set at two); it only stops ordering rows.

**Do not import `ladder_index` into either function.** The tiebreak is binary — `Broken` or not. The full ladder is Codex's ordering, and applying it to Nate's queue is rejected on #32. `startable()` (`funnel.py:358`) is not touched.

## How to tell it worked

Tests in `tests/`, fixtures only, no network — in the style of `test_ordering.py`:

- a `Broken` item at a shallower gate does **not** outrank an item at a deeper gate
- a `Broken` item does outrank a same-gate item that has waited longer
- a ticket whose **parent** is `Broken` sorts first (the `by_ref` inheritance path)
- Ideas sort `Broken` first, then oldest
- an unlabelled idea is no longer demoted below labelled ones

## Scope

Stops at the sort keys. Rendering `Class` in these lists is a sibling ticket; capture always labelling is another. Do not touch `plan.md`, `AGENTS.md`, or the label set.

Risk: standard

nateprich:
Closes #44.

Updated awaiting_decision to materialize a parent map and sort by gate depth, blocked status, inherited Broken class, age, repo, and number. Updated ideas to sort Broken first then age, ignoring needs-shaping for ordering. Added fixture coverage for the requested cases.

Deliberately did not change startable, ladder_index, Status/Class values, label definitions, plan.md, or AGENTS.md. Verification: python3 -m pytest -q (280 passed).


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "8ddca6785c0dfc8ca4e88f23ecfc88654606e791",
  "reviewed_at": "2026-09-07T16:39:41.494352+00:00",
  "verdict": "approved"
}
```

Issue #151, opened 2026-09-08
Unmarked origin resolves to Nate, with backlog fixture test


nateprich:
Sub-issue of #90 (Ticket shape item 4). Depends on #150 — the default lives in the same rule #150 extends; if #150 has landed, amend its term, otherwise treat absent origin as not-agent-shapeable and say so in the PR.

**What.** Items predating any origin marker resolve to Nate's: absent or malformed origin means shaped with him (fail closed, the same choice as #31 for unattributed comments). Add a fixture test over the existing twelve pre-marker ideas confirming none is treated as agent-shapeable.

**Read first.** #90 body (existing-backlog section, unmarked-fails-toward-Nate precedent), #31, #150 as landed.

**Acceptance.**
- Absent origin resolves to Nate (not agent-shapeable); malformed origin likewise.
- Fixture test enumerating the twelve existing backlog ideas, asserting agent-shapeable is false for every one.

**Rejected (do not reintroduce).** Backfilling origin onto the twelve by inference — the defect #31 exists for.

Risk: standard

Issue #150, opened 2026-09-08
Self-approval rule keys on origin as well as class


nateprich:
Sub-issue of #90 (Ticket shape item 3). Depends on #84 — #84 is OPEN and adds another term to the same rule; whichever of the two lands second must amend the first rather than assume its own shape. Read #84's current code before starting.

**What.** #59's self-approval rule keys on class AND origin: Nate raised it implies shaped with him whatever its class; an agent observed it implies the existing class rule applies. #77 and #80 own the condition functions — add a term to them, not a second gate. Effective origin for the rule is: explicit origin, else override, else unmarked-means-Nate (see the unmarked ticket, #151 — do not reimplement its default here beyond treating absent origin as not-agent-shapeable).

**Read first.** #90 body (the thing, the #84-independence note, rejected: keying on Class alone and overloading Class with a sixth value), #59, #77, #80, #84 as landed.

**Acceptance.**
- Nate-raised Improve/Maintenance/Broken requires Nate (not agent-shapeable).
- Agent-observed item follows the existing class rule unchanged.
- Tests for both, plus a test that the #84 term (as landed, or absent) still holds — no second gate, one combined condition.

**Rejected (do not reintroduce).** Keying on Class alone (the approved #59 rule this issue rejects); overloading Class with a sixth value meaning 'mine' (Class is the engineering queue's sort key in startable()); an explicit opt-in flag only, returning the default to Nate-in-the-loop on everything.

Risk: escalated — gate, who may be shaped unattended

Issue #148, opened 2026-09-08
funnel capture records origin with no default


nateprich:
Sub-issue of #90 (Ticket shape item 1). Depends on #61 for the voice convention — #61 is closed, so build on its landed implementation.

**What.** `funnel capture` records origin at capture using #31's voice convention (the `nate-relayed` / `agent` distinction; see #31 body and #61's parser/writer). Origin states whether the capturing agent was told (Nate raised it) or observed it itself. There is no default: capture refuses without an explicit origin, fail closed.

Follow the house style #61 set (marker constant plus parser, \--voice-style flag; say which flag and marker in the PR). The parser must not confuse the origin marker with REVIEW_MARKER.

**Read first.** #90 body (origin recorded-not-inferred, marker follows #31, plan.md:124 closes the set at two — no new label), #31, #61.

**Acceptance.**
- `funnel capture` without an origin refuses.
- Capture with each origin records it retrievably (marker or field, per #61's convention).
- A test for the refusal and for both origins recorded.

**Rejected (do not reintroduce).** Backfilling origin onto the existing twelve ideas — inference about authorship is the defect #31 exists for. Unmarked items are T4's job, not this ticket's.

Risk: standard

PR #181, opened 2026-09-08
Add asymmetric origin override parser

### LINKED ISSUE #149: Origin override marker with asymmetric permission
Sub-issue of #90 (Ticket shape item 2).

**What.** The origin override marker and its parser. Origin is the default and Nate may override it in either direction (hand over one of his, or pull back an agent-found item). Permission is asymmetric and load-bearing: toward Nate by anyone, toward agents by Nate alone. An agent marking an item 'no discussion needed' would be authorising its own unattended shaping — the self-authorising shape #84, #89 and #90 are each a version of.

Follow the #61 house style (marker constant plus parser; distinguish on the HTML comment exactly as the review/provenance parsers do). No new label (plan.md:124).

**Read first.** #90 body (the override section and the regression note), #31, #61.

**Acceptance.**
- Nate setting the override toward agents is accepted.
- Any agent (any voice) setting the override toward Nate is accepted.
- A non-Nate voice setting the override toward agents is refused (fail closed), with a fixture test for each direction.
- Parser tests: override marker is never confused with REVIEW_MARKER or the provenance/origin marker.

**Rejected (do not reintroduce).** A symmetric override — however natural it looks, it hands an agent the ability to authorise its own unattended shaping. Origin alone with no override (rejected on Nate's call: origin is a proxy and will be wrong at the edges).

Risk: escalated — authorisation, who may authorise unattended shaping

nateprich:
Closes #149.

## What changed

- Added the distinct command-center-origin-override HTML marker and its accepted targets: nate or agents.
- Added a parser that accepts every override toward Nate, but accepts an override toward agents only when the same text carries a valid nate-direct or nate-relayed provenance voice.
- Added fixtures for all three existing voices, missing and malformed provenance, invalid targets, and isolation from the review, provenance, and forthcoming origin markers.

The toward-Nate path deliberately accepts an unattributed marker: it only adds Nate to shaping and therefore fails in the safe direction. The toward-agents path fails closed without valid Nate provenance, preventing an agent voice from authorising its own unattended shaping.

## Branch continuity

This is a fresh ticket/149 branch from current main. The prior-run lookup matched only this run's own setup activity; there was no stranded implementation or existing remote branch to continue.

## Deliberately not done

- No CLI writer was added; this ticket asks for the marker and parser only.
- Capture origin, unmarked-origin fallback, and the self-approval condition remain with sibling tickets #148, #151, and #150.
- No label was added or changed.

## Verification

- python3 -m pytest -q: 374 passed
- git diff --check: clean


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "ff497a2a47e7073c335dac796f07b120b40ac67a",
  "reviewed_at": "2026-09-08T18:10:14.896887+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-08T18:10:17.617979+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #84, opened 2026-09-07
#59's primary brake is self-certified — `plan_needs_nate` reads the plan the agent wrote


nateprich:
Shaped 2026-09-07 with Nate, within the hour of the defect being found.

## The defect

#59 lets an agent shape a `Broken`/`Maintenance`/`Improve` item to `Ready` without Nate
when two conditions hold: the plan's "Needs you" section is empty, and the plan carries no
escalated risk.

**Condition 1 is self-certification.** #77 builds `plan_needs_nate()` to read — per #59's
own plan — "the plan's own sections and body". The agent writes the section; a function
then reads that section to decide whether the agent may skip Nate. The check asks *did you
say you need Nate?* An agent that judges wrongly, or omits a question it never recognised
as his, passes its own test.

That is #31's failure shape at a gate: an agent's unverifiable claim about its own work,
treated as fact by the thing that acts on it.

**Condition 2 does not cover for it, and that was measured rather than argued.** On
2026-09-07 both #82 and #83 were passed through `escalation_reasons()`; both returned
`NONE` — standard tier. #83 contains a genuinely Nate-only question: whether an
incompletely onboarded repo may be refused entry, where `plan.md:417` makes membership
itself a gate and `plan.md:740` keeps applying topics and transferring issues with him.

`ESCALATION_PATTERNS` looks for credentials, authorisation, migrations, destructive
operations and concurrency. **Authority-shaped content is not exposure-shaped content**,
and the design as approved assumes they are the same thing.

## The thing

**Verify the "Needs you" section against the plan it sits in.** Nate's call, 2026-09-07.

An empty "Needs you" is refused when the plan's own content does something only he may
authorise. The agent still writes the section, but a claim of "nothing needed" now has to
survive a reading of the plan around it rather than being taken on trust.

Signals that contradict an empty section — the starting set, meant to grow:

- the plan cites `plan.md` or `AGENTS.md` on a **gate**, on **membership**, or on **who
  may write** something
- it changes **what an agent may do unattended**
- it changes a **gate's question, its answer, or who answers it**
- it changes **who may set a field** that other rules act on

**A flagged plan stops and returns to Nate.** He chose this direction knowingly: the check
is pattern-matching on prose, so it will have false positives, and a false positive costs
him one look at something he did not need to see. That is the safe direction and the
friction is accepted rather than discovered later.

### The consequence for how plans are written

Verifying a section against content requires the section to make a claim. **An empty
[...]

Issue #80, opened 2026-09-07
`funnel shaped` advances a qualifying plan to `Ready` itself (self-approval)


nateprich:
Part of #59, and it is the handover itself. Read #59's Rejected list first — "Nate confirms the class", "class earned by mechanical evidence" and "shadow mode before handover" are all rejected. The trigger is the plan's own "Needs Nate" section plus no escalated risk plus a qualifying class, and nothing else.

Depends on #76 (the re-tuned patterns — #59 is explicit that the handover never runs on the old ones) and #77 (these checks). Do not start while either is open.

**Also depends on #32**, which owns the Class-at-entry policy this amends ("`Broken` settable at entry by Nate alone"). Check its state before starting: if it has landed, write this as the deliberate amendment #59 calls it; if it is still in Building, do not assume its rules — say so in a comment here and leave this blocked for Nate to reconcile. Note `skills/shape`'s 2026-09-07 capture-classing rule already moved part of that policy, so the reconciliation may be mostly a matter of recording it.

## What to build

- In `funnel shaped` (`funnel.py:2032`): once the plan body is written, evaluate all three conditions — `Class` is one of `Broken`/`Maintenance`/`Improve`; `plan_needs_nate` is false; `plan_is_escalated` is false. All three hold → Status becomes `Ready`, not `Shaped`, and a comment on the issue records the self-approval **marked as the agent's own decision**, naming which checks passed. Give the marker line a stable machine-readable prefix (e.g. `Self-approved:`) — the brief-visibility sibling surfaces it. Any condition failing → today's behaviour unchanged: `Shaped`, waiting on Nate.
- `New` and `Replace` never qualify, whatever the plan says — building new and replacing what works are scope decisions and stay his, the same line #56 draws on the accept side.
- Class unset never qualifies — unset must not silently route a project past him (`plan.md:491`); `needs_class` already nags about the field itself.
- Rewrite the `ANSWERS` guardrail comment near `cmd_shaped` rather than deleting it: no agent may *hand-write* `Ready`, and the funnel writes it only through these checks. The boundary gets narrower and truer, not removed.
- `plan.md`'s surfaces section (2026-09-05 revision) already says `funnel shaped` advances a plan with no open question to `Ready` itself — with **no class condition**, so the record currently over-permits what the code does. Amend that paragraph in this PR, citing #59's 2026-09-06 decision, to the class-qualified rule.
[...]

Issue #77, opened 2026-09-07
`plan_needs_nate()` and `plan_is_escalated()` — the checks self-approval turns on


nateprich:
Part of #59. Two pure functions and nothing else — the shaping path that acts on them is the self-approval sibling, and it must not land before these exist.

## What to build

- `plan_needs_nate(plan_body)` → true when the plan's "Needs Nate" section holds anything other than an explicit nothing. Plans here end that section with "Nothing." (#59's own does); treat the bare nothings — "Nothing", "Nothing.", whitespace — as empty, and anything else (a question, an open point, a deferral) as needing him. Accept both heading spellings in the wild: "Needs Nate" and "Needs you" (`plan.md` uses each in different sections).
- A plan with **no such section at all** needs Nate. Missing is not empty: a plan that never justified skipping him has not earned skipping him — unset must never default to the most permissive reading (`plan.md:491`, the same rule `needs_class` enforces).
- `plan_is_escalated(plan_body)` → the (re-tuned) `ESCALATION_PATTERNS` machinery applied to the whole plan body, reasons returned in the `escalation_reasons` style. Whole body on purpose: a risky plan does not become safe by keeping the word "credentials" out of its headings.
- Fixtures: a needs-Nate plan; a "Nothing." plan; a sectionless plan; a plan whose body trips the re-tuned patterns; this repo's lock-and-gate vocabulary used as subject matter.

## Sequencing

Naturally lands after the patterns re-tune sibling, so its escalated fixtures reflect the new boundary — but no hard dependency: these tests exercise the wrappers' parsing, not the patterns' tuning.

## Scope

Pure functions plus fixtures. No change to `funnel shaped`, no gate writes, no skill edits.

Risk: standard

nateprich:
Closing: the work is on `main` and verified.

PR #249 (branch `ticket/77`) merged at 14:54 PDT carrying **no closing-issue
link**, so this stayed open and stayed claimable. Codex then re-claimed it and
re-verified already-merged work on the runs at 15:28, 15:33, 15:40 and 15:43,
each finishing `done` having produced nothing.

Verified now against `origin/main`: `plan_needs_nate()` at funnel.py:623 and
`plan_is_escalated()` at funnel.py:639, with `tests/test_plan_checks.py` from
the same PR. 454 tests pass.

Third instance today of the same failure — after #156 (PR #232) and #169
(PR #234). #236 tracks the cause: `funnel merge` runs `gh pr merge` and never
[...]

PR #182, opened 2026-09-08
Key self-approval on origin and class

### LINKED ISSUE #150: Self-approval rule keys on origin as well as class

nateprich:
Closes #150.

## What changed

- Added one `self_approval_eligible` predicate combining the existing class rule with effective origin, the Needs-Nate result, and escalation.
- Added `effective_shape_owner` to normalize capture provenance voices and an already-authorized origin override without duplicating either marker parser.
- Made missing, malformed, or untrusted origin fail closed: it is not agent-shapeable.
- Added fixtures proving Nate-raised Broken/Maintenance/Improve items still require Nate, while agent-observed items follow the existing class rule unchanged.
- Added coverage for both override directions and for #84's landed authority signal remaining part of the same combined condition.

## Branch continuity

This is a fresh `ticket/150` branch from current `main`. The prior-run lookup found only the setup activity from this run, and no remote `ticket/150` branch existed.

## Dependency reconciliation

#84's current verifier is already on `main` as `needs_nate_signals`; this PR feeds its result into the same eligibility predicate rather than adding a second gate.

#149's origin-override parser is still open in PR #181. This PR deliberately accepts only its normalized target (`nate` or `agents`) and does not duplicate the marker or authorization parser. A merge-tree check against `ticket/149` reports no conflict.

#148, #151, #77, and #80 have not landed. Their responsibilities remain separate: capture writes origin, the backlog default resolves unmarked origin, the plan checks produce the two booleans, and the shaping command performs the transition. #80 can call this one predicate when those inputs exist.

## Deliberately not done

- No second gate and no Status write were added; #80 owns the live shaping transition.
- No origin or override marker parser was added; #148 and #149 own those.
- No backlog default or backfill was added; #151 owns the fail-closed legacy fixture.
- No `Class`, `Status`, label, or issue metadata was changed.

## Verification

- `python3 -m pytest -q`: 382 passed
- `git diff --check origin/main...HEAD`: clean
- merge-tree check with `origin/ticket/149`: no conflict markers


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "b04202dd99eba5e00e0a52ae1a8548d70fdbac51",
  "reviewed_at": "2026-09-08T17:10:30.172521+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-08T17:10:33.198965+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #243, opened 2026-09-08
Propagate dependency priority to blockers

### LINKED ISSUE #173: 129/6: a blocker inherits the rank of the work it blocks, derived not written
Parent: #129. Depends on #153 (the `blocked_by` reader) — the graph must be loadable before rank can be derived from it.

What: in `startable()`, a ticket's effective rank becomes the best rank of itself and everything it transitively blocks. A blocker is never offered later than the work waiting on it. Derived at sort time from the native `blocked_by` graph #153 loads — **no write to `Class`, no stored field, nothing to reverse when the dependency clears.**

Why this and not the obvious thing: reclassifying the prerequisite to `Broken` was considered on 2026-09-08 and rejected by Nate. Two reasons, both measured. Within a class the sort key is `(status_since, repo, number)` and **0 of 91 tickets carry a `status_since`**, so it collapses to issue number — promoting the class orders correctly only when the prerequisite happens to have the lower number. #77 (position 30, `Improve`, under #59) sorting ahead of #108 (position 1, `Broken`, under #84) would have been luck, not construction. And `Class` feeds `upkeep_share`, the portfolio signal the funnel is told to flag and never tune; relabelling improvements as defects inflates it permanently with no reverse ratchet.

The failure this prevents, in full: on 2026-09-08 #108 sat at position 1 needing #77 at position 30. `funnel next` returned #108 to every run, each read the prose dependency, correctly declined, recorded `errored`, and was handed it again — **26 consecutive Codex runs over six hours, 41 that week**, with Muse idle behind it for want of anything to review. Its sibling #154 stops the wedge by excluding a blocked ticket; only this one makes the blocker actually happen sooner, so a `Broken` project stops waiting on a low-ranked prerequisite.

Accept: a `Broken` ticket blocked by an `Improve` ticket puts the `Improve` one first, **including when the blocker's issue number is higher** — that case is the whole point and must be an explicit test. Transitive chains lift the deepest blocker. A cycle terminates and does not recurse forever. A ticket blocking nothing keeps its own rank. `Class` values in the Project are unchanged after sorting, and `upkeep_share` is unaffected — assert both.

Do not: write `Class`, or any Project field (the rank is derived, never stored). Add a label — `plan.md` closes the set at two. Parse a `Depends:` body line — rejected on #153, a typo reads as no dependency. Reorder Nate's queue: `awaiting_decision()` is untouched, its `Broken`-first tiebreak is his decision recorded on #32 and this is Codex's ordering, not his.

Risk: standard

nateprich:
Closes #173

## What changed
- Reverse the loaded native `blocked_by` facts into a cycle-safe descendant graph.
- Rank each startable ticket by the best ladder class of itself and everything it transitively blocks.
- Keep dependency ancestors ahead on equal effective rank with a deterministic descendant-count tie-break.
- Leave `Class`, Project fields, `awaiting_decision()`, and `upkeep_share` unchanged.

## Verification
- `python3 -m pytest -q` — 430 passed.
- Task-local-cache `python3 -m compileall -q funnel.py heartbeat.py prior_run.py usage.py` — passed.
- `git diff --check` — passed.
- A live `python3 funnel.py brief` read produced no output within the 30-second tool window, so live verification is not claimed.

No plan, Status, Class, label, or dependency writes were made.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "1efd04d18570ebfd261c40f6de997dbe0b8f99fe",
  "note": "Meets #173: effective rank derived at sort time from native blocked_by graph, no Class/Project writes, higher-numbered blocker case tested, transitive + cycle + no-mutation tests present, awaiting_decision untouched. Touches funnel.py + tests/test_ordering.py only.",
  "reviewed_at": "2026-09-08T21:18:44.804814+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-08T21:18:47.133963+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #129, opened 2026-09-08
Ticket dependencies live in prose, so the queue hands out work that cannot be done


nateprich:
Shaped 2026-09-07 with Nate, hours after it started costing runs.

## The defect

Ticket dependencies are written in prose. `startable()` cannot read prose, so it hands
out work whose prerequisites have not landed.

**Measured 2026-09-07: 15 of 40 startable tickets carried a prose dependency** — #48,
#57, #58, #61, #62, #63, #67, #73, #74, #81, #108, #109, #111, #112, #113. Breakdown
records them faithfully in ticket bodies and coverage comments. Nothing reads them.

Two costs, and the second is worse than the first.

**Runs burn their opening and stop.** Codex was handed #73, whose body says *"Depends on
#71 and on #35 landing"* and *"if its rewrite has not landed, leave this ticket for a
later run"*. It read that, correctly declined, made no changes — and was handed the same
ticket again. It recorded `errored` both times, because `OUTCOMES` has no way to say "I
correctly declined", and `.github/scripts/watchdog.py` alarms at three errors in a week.
**Doing the right thing was trending toward a fault alarm.**

**Dependent tickets get worked in parallel and collide.** #66, #67 and #68 all rewrite
the watchdog's silence logic, and #67's body says it depends on #66. All three were
worked at once. All three PRs — #121, #122, #123 — went `DIRTY`. Muse approved #121 and
the merge gate refused it: *"not mergeable, merge commit cannot be cleanly created."*
Merging any one of them makes the others conflict harder, so the rework loop feeds
itself. That is a much more expensive failure than a wasted poll, and it looks like
progress while it happens.

## The thing

**GitHub models this natively, and it is already available on this repository.** Verified
2026-09-07: `GET /repos/{owner}/{repo}/issues/{n}/dependencies/blocked_by` and
`.../blocking` both answer, and a `POST` with `issue_id` creates the relationship. #67 is
now recorded as blocked by #66 — the first real one, written during shaping.

**`startable()` excludes a ticket whose blocker is still open.** Nate's call,
2026-09-07: silently, the same way it already excludes a ticket whose work is in an open
PR. No label, no comment, no decision. The graph is the truth and the queue reads it
every time, so it self-corrects the moment the blocker closes.

That is the whole difference from what we did by hand today. Five `blocked` labels were
applied manually, and every one of them needs a person to remember to remove it.

**Breakdown writes the relationship**, at the moment it already knows. It writes the same
[...]

PR #269, opened 2026-09-09
Make doctor item consistency report real contradictions

nateprich:
Closes #266.

## What changed

- Exempt child tickets from the project-only CLOSED/Status contradiction.
- Add a pure merged-PR contradiction using `ticket/<n>` branch refs passed in from the caller.
- Have `funnel doctor` scan merged PR heads once per member repository, bounded to the newest 100 rows (requesting one extra row to detect truncation). A truncated scan is named in the stranded finding.
- Fail the item-consistency check closed if the merged-PR scan cannot be read.

## Deliberately not changed

- No ticket or parent is closed automatically.
- No Project Status/Class fields are written.
- `doctor` does not call the per-ticket `ticket_pr_facts()` lookup.
- The inverse case (closed ticket whose PR never merged) remains excluded because it would flag legitimate `not_planned` tickets.

## Verification

- `python3 -m pytest -q tests/test_doctor.py` — 47 passed
- `python3 -m pytest -q --ignore=tests/test_automation_drift.py` — 467 passed
- task-local `compileall` and `git diff --check` passed
- live `python3 funnel.py doctor` produced the five expected real `needs-shaping` contradictions and no merged-ticket findings

The full local suite is 474 passed with one pre-existing environment-only automation-drift failure: four host automation copies are stale; this PR does not touch routines or automation configuration.

## Continuation

Codex run `0768af5a3db3` found the remote `ticket/266` branch and PR #269 after an overlapping branch push. Continued that existing implementation without resetting or force-pushing, and independently verified the scoped diff. Local verification: 47 focused tests, 467 non-drift tests, 474 full tests with the pre-existing automation-copy drift failure, task-local compileall, `git diff --check`, and live `funnel doctor` showing the five expected real `needs-shaping` contradictions with no merged-ticket findings.

Issue #266, opened 2026-09-09
`doctor` item consistency reports only real contradictions


nateprich:
Parent: #254.

## What

Two changes to `item_consistency_findings()` (funnel.py:1930), shipped as one ticket because both edit the same function — the plan's own stated reason for folding #252 in ("splitting them is two runs and a conflict").

1. **Exempt tickets from the `CLOSED` rule.** Give it the same `item.parent` guard `needs_class()` already opens with, rather than inventing a second way to say "this is a ticket". Projects must still be caught: a closed project whose `Status` is not `Done`/`Parked` is a real contradiction.
2. **Add the stranded-ticket finding.** Report an open ticket whose `ticket/<n>` PR is already merged. Detect it by listing merged PRs once and intersecting the `ticket/<n>` head branches with open tickets — one request regardless of backlog size.

## Constraints from the plan

- **Report, never auto-close.** Do not touch parent `Status` or state. A diff that does reintroduces #55, which is blocked precisely because auto-closing removes the accept gate.
- **Do not use `closedByPullRequestsReferences`.** Measured 2026-09-08: it returns empty for #77 (stranded by a PR with no `Closes` link) and `[185]` for #140 (which had one) — inverted from what this check needs. `linkedBranches` is empty for both, because `ticket/<n>` is a local convention.
- **Keep the finding function pure over `Item`s.** Pass the PR facts in, per `startable()`'s stated convention.
- **Do not make `doctor` load `ticket_pr_facts()`** — ~78 requests per run, the cost shape #237 exists to stop paying.

## Judgement left to this ticket

How far back to scan merged PRs. Pick a bound, state it in the code, and make it visible in the finding when the scan is truncated. An unbounded scan re-introduces a cost that grows with history.

## Accept

Fixture tests: a closed ticket produces no finding; a closed project whose `Status` is not `Done`/`Parked` still does; an open ticket whose `ticket/<n>` PR is merged is reported; one whose PR is unmerged is not; a truncated scan is visible in the output. Against the live Project, `funnel doctor` item-consistency findings drop from 63 to the 5 real `needs-shaping` ones.

Risk: standard

Issue #254, opened 2026-09-08
`doctor` item consistency flags every closed ticket, because tickets carry no Status by design — 58 of 63 findings are false


nateprich:
## What this is

`item_consistency_findings()` (funnel.py:1930), landed by #247 at 14:34 on
2026-09-08, applies its `CLOSED` rule to every item:

```python
if item.state == "CLOSED" and item.status not in ("Done", "Parked"):
```

Tickets carry no `Status` by design, so every closed ticket trips it. Measured
against the live Project: **63 findings, 58 of them false** (92%). The five real
ones — `needs-shaping` left on #29, #30, #65, #75 and #114 — are invisible inside
the noise, and `doctor` prints one line per item.

This also absorbs #252, closed as superseded on Nate's call: report an open
ticket whose PR is already merged. Both changes touch the same function, so
splitting them is two runs and a conflict.

**The work:**

1. **Exempt tickets from the `CLOSED` rule.** A ticket has a parent; give the
   rule the same `item.parent` guard `needs_class()` already uses. Projects must
   still be caught — a closed project whose Status is not `Done`/`Parked` is a
   real contradiction.
2. **Add the stranded-ticket finding.** Report an open ticket whose `ticket/<n>`
   PR is already merged. **Report, never auto-close.**

## Decided from precedent

- **Tickets carry no `Status`, so a closed ticket with none is correct.**
  `plan.md`: *"the project-level issue carries `Status` and `Class`. Its
  sub-issues are tickets and carry neither."* `skills/breakdown/SKILL.md` repeats
  it: *"Do not set `Status` or `Class` on tickets. Blank fields on a child are
  correct."* So this is a deduction, not a judgement call.
- **The exemption already has a shape in this file.** `needs_class()` opens with
  `if item.state == "CLOSED" or item.parent: return False`. Use the same guard
  rather than inventing a second way to say "this is a ticket".
- **Report, never auto-close.** #252's plan said so, and #55 — auto-closing a
  parent — stays blocked because it removes the accept gate for
  `Broken`/`Maintenance`/`Improve`. Do not touch parent `Status` or state; a
  reviewer should treat any diff that does as reintroducing #55.
- **`doctor` is where this family of contradiction lands**, per #53, #132 and
  #218.
- **Detection must not cost one call per ticket.** ff35327 removed exactly that
  shape from `load_items()` after it exhausted the API budget and took `funnel
  brief` down; #237 tracks the general problem. `ticket_pr_facts()`
[...]

PR #271, opened 2026-09-09
Right-size ITEM_QUERY's connections to cut GraphQL cost 4.9x

nateprich:
`ITEM_QUERY` requested **7,300 nodes per 50-item page**. GitHub charges `ceil(nodes/100)`, so that was 73 points per page and **365 points for a single `load_items()`** across the 202-item board. The budget is 5,000/hour and `main()` loads once per process ([funnel.py:3868](https://github.com/nateprich-projects/command-center/blob/main/funnel.py#L3868)), so **13 funnel commands an hour exhausted it**. Roughly 52 happen — the funnel was down about three quarters of every hour, which is the outage observed on 2026-09-08.

| Connection | Before | After | Why |
|---|---|---|---|
| `timelineItems` | `last: 60` | `last: 10` | Only the most recent status change is read |
| `blockedBy` | `first: 50` | `first: 10` | Items carry 1–2 blockers |
| `labels` | `first: 25` | `first: 5` | The label set is closed at two by convention |
| `assignees` | `first: 10` | `first: 3` | 0–1 |

7,300 → 1,450 nodes per page: **365 → 75 points per load, 4.9x cheaper.** 66 affordable loads an hour against the ~52 the routines need.

The bounds keep deliberate headroom rather than hugging observed maximums. `blockedBy` stays at 10 against a live maximum of 2, because a missed blocker hands out unstartable work — the failure #129 fixed. `timelineItems` stays at 10 although an item still at its current status cannot have had a later status change, so its entry event is the most recent one; the buffer covers events from other projects, which interleave before the project-number filter is applied.

**No logic changed.** Full offline suite: 457 passed, with `test_automation_drift` failing identically before and after (pre-existing).

**Process note:** this was applied to Nate's working tree before review, on his explicit instruction in a /funnel session on 2026-09-08, because the gates it would normally pass through run *through* the funnel — which was down 74% of the time until this landed. He decided; I typed.


nateprich:
Closing this: **its premise was wrong, and measurement disproved it.**

The change was justified by a modelled cost of 365 points per `load_items()`, derived from GitHub's documented worst-case formula `ceil(nodes_requested / 100)`. GitHub actually charges on nodes **returned**, not requested. Measured directly against the live Project on 2026-09-08, with `rateLimit { cost }` in the response:

| Query shape | Measured cost |
|---|---|
| Original (`labels 25`, `blockedBy 50`, `timelineItems 60`) | **2 points/page** |
| This PR (`labels 5`, `blockedBy 10`, `timelineItems 10`) | **2 points/page** |

So the connection right-sizing saves nothing measurable, and the "4.9x cheaper" claim in the commit message and description is false.

**What is actually expensive**, measured the same way: a full `funnel brief` costs **110 points**, and **93 of them (85%) are `ticket_pr_facts()`** — `_ticket_pr` (funnel.py:3167) runs one `gh pr list --head ticket/<n>` per open ticket, 68 separate calls. It is an N+1, not a fat query. `load_items()` is ~12 points and `member_repos()` is 5.

The fix for that already exists in this repo: #266 (merged as #269) implements the one-call technique #254's plan specified — list PRs once, take the `ticket/<n>` head branches, intersect with open tickets. `ticket_pr_facts` should adopt it, which takes a brief from 110 to roughly 20 points.

Captured separately. Nothing here is worth keeping, and leaving it would put a false cost claim in the history as settled fact.

Nate approved landing this before review during what looked like an outage, on my analysis; the analysis was wrong, and he directed the revert once it was measured. His working tree has been returned to `main`.

Issue #210, opened 2026-09-08
Routine work step: on declining a ticket ask again excluding it, up to three times, then finish as skipped-blocked naming each


nateprich:
Parent: #177 — a run that declines the head of the queue should try the next one, not go home.

## What to build (from the plan)

Each routine's work step: on declining a ticket, ask again excluding what it just declined, up to three candidates total, then finish.

- Uses the `--not` exclusion from #209 to re-ask; no other queue mechanism.
- A run declining three in a row stops; it records `skipped-blocked` (defined in #175) naming EACH ticket declined, so the run note says what it stepped over rather than only what it stopped on.
- Routines are pasted copies: include the automation re-sync and regenerated `--routine-sha` literals in the SAME change (as #171 and #175 also do).
- Must NOT implement the plan's rejected alternatives: no persistent decline memory with a TTL, no reordering/demoting declined tickets.

## Acceptance

Fixtures only, no network:
- A run declining three in a row stops and its heartbeat names all three.
- Replay 2026-09-08: given a queue headed by an unworkable #108 with workable tickets behind it, a run reaches a workable one instead of stopping.

## Dependencies

Depends on #209 (this re-ask needs the `--not` exclusion) and on #175 (the `skipped-blocked` outcome this records must exist first). Do not start this ticket while #209 or #175 is unlanded; the queue cannot read this prose (tracked as #129), so also check before starting.

Risk: standard

PR #278, opened 2026-09-09
Add per-call exclusions to funnel next

### LINKED ISSUE #209: funnel next --not <ref>: exclude already-declined refs, pure filter over the existing queue
Parent: #177 — a run that declines the head of the queue should try the next one, not go home.

## What to build (from the plan)

Give `funnel next` a way to exclude refs the caller has already declined this run — a repeatable `--not <ref>` or equivalent.

- Purely a filter over the existing queue; no ordering change.
- No new persisted state: per-call filter only, nothing stored on the ticket or the queue.
- Must NOT implement the plan's rejected alternatives: no recorded decline with a TTL, no reordering/demoting declined tickets.

## Acceptance

Fixtures only, no network:
- `next` with an exclusion returns the following candidate.
- With every candidate excluded it returns nothing.
- Ordering of the non-excluded candidates is unchanged.

## Out of scope

Routine retry loop and heartbeat note — that is the follow-up ticket. Re-sync of automations is with the follow-up ticket in the same change as the routine edit (per the plan, routines are pasted copies).

Depends on: nothing.

Risk: standard

nateprich:
## What changed

- Add repeatable `funnel next --not <ref>` arguments.
- Filter excluded refs after the existing ordered `startable()` queue is built.
- Apply the same filter in tier-specific selection so exclusions cannot be bypassed.
- Add fixture-only coverage for next-candidate selection, all-excluded queues, unchanged ordering, and repeatable CLI flags.

## Deliberately not changed

- No persisted decline state, queue reordering, or TTL.
- No routine retry loop or automation re-sync; those are scoped to #210.
- No Status/Class changes.

## Verification

- `python3 -m pytest -q tests/test_ordering.py` — 87 passed.
- `python3 -m pytest -q` — 479 passed; one known pre-existing failure in `tests/test_automation_drift.py` because four host automation copies are stale.
- Task-local `compileall` and `git diff --check` passed.

Closes #209.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "47266f6f368ca05bb9cbd93dc0cdd1f3480fd640",
  "reviewed_at": "2026-09-09T02:25:49.211313+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T02:25:50.432397+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #283, opened 2026-09-09
Fix moved checkout doctor diagnostics

### LINKED ISSUE #260: Replace unsatisfiable install-symlink doctor check with self-retiring staleness check
Parent: #238.

What: In funnel.py `_expected_links()`, drop the self-referential `~/.claude/command-center` -> checkout entry (keep the real skill links `~/.claude/skills/<name>` -> `checkout/skills/<name>`); add a self-retiring staleness check that reports `~/.claude/command-center` behind `origin/main` only while `routines/*.md` still invokes funnel.py from that old absolute path (grep, same fact #205 greps for); correct AGENTS.md lines 189-193 to the replacement prose keeping the hazard and fixing the mechanism; fix the remaining install-symlink suggestion so it never suggests a command install.sh refuses to run (narrow to skill links it can repair, or plain manual steps). First confirm by reading `scripts/install.sh` whether the skill links are still ones it can repair, so the suggestion is true rather than assumed.

Accept: `funnel doctor` no longer reports the unsatisfiable symlink line on the moved checkout; while routines still reference the old path the staleness check is live (today all 22 invocations match), and with zero matches it goes quiet; AGENTS.md prose matches the replacement; running the suggested fix for any remaining symlink error succeeds or gives correct manual steps. Do not add a doctor check on the run-clone path (belongs to #171 work).

Risk: standard

nateprich:
Closes #260.

## What changed

- Removed the self-referential `~/.claude/command-center` entry from the install-link check; statusline and discovered skill links remain covered.
- Added a read-only, self-retiring checkout-staleness check. It compares the canonical checkout with `origin/main` only while `routines/*.md` still contains legacy `funnel.py` invocations, and goes silent after the routine migration.
- Made symlink repair output distinguish installer-repairable links from real files or missing checkout targets.
- Made `scripts/install.sh` idempotent when the canonical checkout is itself `~/.claude/command-center`, so the suggested repair can actually reach statusline and skill links.
- Corrected the `AGENTS.md` explanation of the working-tree hazard.

## Branch continuity

Continued the existing remote `ticket/260` branch and PR #283 after verifying its two commits; no reset or force-push was used.

## Deliberate boundaries

This PR does not change `routines/*.md`, `skills/funnel/SKILL.md`, the future run-clone path, or Project `Status`/`Class` behavior. It does not fetch, write, or otherwise modify the canonical checkout from the doctor check.

## Verification

- `python3 -m pytest -q tests/test_doctor.py` — 51 passed.
- `bash -n scripts/install.sh` — passed.
- `PYTHONPYCACHEPREFIX=/private/tmp/command-center-260-pycache python3 -m compileall -q funnel.py` — passed.
- `git diff --check` — passed.
- Full `python3 -m pytest -q` — 491 passed; the existing `tests/test_automation_drift.py` failure remains for four host Codex automations that are stale relative to the routine.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "b0826557fbd14d73f6d3a204c6f84559e8f6e844",
  "reviewed_at": "2026-09-09T03:19:27.013355+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T03:19:28.991920+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #299, opened 2026-09-09
Record the measured before/after PR-scan cost in LEARNINGS.md


nateprich:
Parent: #272.

Depends on #292 and #295 — the instrumentation and its per-run logging are what make this a measurement rather than a projection — and on #297, which is the change being measured.

**This ticket exists in this order deliberately.** #271 is the worked example: an unmeasured cost model (365 points per load, from GitHub's documented formula) survived to a merged-intent PR before anyone measured the real 2 points per page. Writing "roughly 18 points" here before the instrumentation exists would repeat exactly that.

What: a `LEARNINGS.md` entry, labelled `measured`, carrying:

- the before/after point counts for a full `funnel brief`, read from `rateLimit { cost }` rather than modelled — the "before" baseline measured 2026-09-08 was 110 points total, of which `ticket_pr_facts()` was 93 (85%), `load_items()` ~12 and `member_repos()` 5
- that `reviews` is available from a PR list query, so no per-ticket round trip is needed to read it

If the measured after-figure differs materially from the projected ~18, record the real number and say so plainly rather than quoting the projection.

Accept: the entry is present, newest first, labelled `measured`, and quotes actual readings taken after #297 landed.

Risk: standard

Issue #292, opened 2026-09-09
Add rateLimit { cost remaining resetAt } to GraphQL queries and surface it from gh_graphql


nateprich:
Parent: #273.

What: add `rateLimit { cost remaining resetAt }` to the queries `gh_graphql` (`funnel.py:2213`) sends, and surface that block to callers. `gh_graphql` currently returns `payload["data"]`, so the block must be exposed **without breaking every caller that indexes into `data`**.

Why this is the highest-value item in the set, and why it is not the outage: on 2026-09-08 a wrong cost model — 365 points per load, from GitHub's documented `ceil(nodes_requested/100)` — survived to a merged-intent PR (#271, closed) before anyone measured the real number, 2 points per page. One `rateLimit { cost }` field would have falsified it in a single call. This prevents a class of error rather than an incident.

**Read cost from `rateLimit { cost }`. Never infer it by differencing `remaining` between calls.** The token is shared — a measurement on 2026-09-08 found 1,826 points spent by something other than the measuring session — so `remaining` moves under the funnel's feet from Nate's own `gh` use and every other agent on the token. A delta attributes other consumers' spend to the funnel.

Accept: a funnel command reports its own per-query cost from the response, and existing callers indexing `data` are unchanged — proved by the suite passing plus one test asserting the block is surfaced.

Risk: standard

Issue #178, opened 2026-09-08
All five Codex automations have drifted, and the 5-minute schedule calls a `funnel next --not` flag that does not exist


nateprich:
# All five Codex automations have drifted, and the busiest one calls a flag that does not exist

**This plan replaces the issue's original subject.** The capture read overlapping
schedules as wasted duplicate openings. That premise is wrong, and investigating it on
2026-09-08 turned up a live defect the original framing hid.

## What is actually true

`scripts/sync_codex_automations.py` generates five Codex Scheduled prompts from
`routines/codex-work.md`, deriving two lanes **from each schedule's own rrule**:

| automation | fires | gate | tier |
|---|---|---|---|
| `command-center-tickets-hourly` | every **5 min**, 24/7 (`BYMINUTE=0,5,…,55`) | `usage.py gate codex --idle` | `--tier standard` |
| the other four | 15-min ticks inside `BYHOUR` windows | `usage.py gate codex` | `--tier escalated` |

`BYHOUR` present means the schedule already encodes "Nate is asleep or at work", so the
idle proxy is omitted and the escalated engine runs. **The overlap is two lanes, not one
lane firing twice.** The original capture's cost argument does not survive this.

## The defect

`python3 scripts/sync_codex_automations.py --check` on 2026-09-08 reports
`5 of 5 automations have drifted from routines/codex-work.md`.

1. **The four escalated copies predate the routine.** Written `12:18:52`; the routine was
   rewritten at `18:15:41` by f534826 ("Record honest blocked declines", #268). They still
   carry pre-#268 decline handling.
2. **The standard copy was hand-edited after the routine changed** — written `19:33:12`,
   after 18:15 — and does not match it. It instructs Codex to run:

   ```
   python3 …/funnel.py next --tier standard --not <declined-ref>
   ```

   **`funnel next` has no `--not` argument.** Measured: `funnel.py: error: unrecognized
   arguments: --not foo/bar#1`. So on the schedule that fires **288 times a day**, every
   decline path invokes a command that cannot parse. The repo's canonical decline
   procedure (`routines/codex-work.md:218`) is different and correct.
3. **The detector exists and nothing runs it.** `--check` reports this precisely, in one
   command, and is invoked by no routine and no test.

## Ticket shape

1. **Reconcile the hand-edited standard copy, then sync all five.** Read what
   `command-center-tickets-hourly` gained at 19:33 and decide, in the issue, whether any
[...]

PR #307, opened 2026-09-09
Replace the per-ticket PR loop with one bounded scan per repo

### LINKED ISSUE #297: Replace the per-ticket PR loop with one bounded scan per member repo
Parent: #272.

What: `ticket_pr_facts` builds a head-branch index from a single `gh pr list --state all` per member repo, instead of calling `_ticket_pr` (`funnel.py:3334`) once per open ticket — 68 separate requests on the current board to read facts that all live in one list. Bound the scan with `MERGED_PR_SCAN_LIMIT` and the `+1` truncation probe (`funnel.py:166`). **Same signature, same return type, same map contract.**

Not a new invention — the technique is in the file twice already: `merged_pr_facts` (`funnel.py:3581`, #269) does the bounded one-call scan with a truncation flag, and `review_queue` (`funnel.py:3364`) already intersects `ticket/<n>` head branches against loaded items. Adopt those; do not design a third variant.

Measured 2026-09-08 with `rateLimit { cost }`: one 100-PR scan with all seven fields and `reviews` nested 20 deep costs **1 point**; one per-ticket `--head` lookup also costs 1. So 68 lookups collapse to 1. `reviews` **does** come back from a list query — verified against the live repo, all seven fields populated, no second round trip.

**Three things this must not get wrong, all in this one diff because shipping any of them separately ships a bug:**

1. **A ticket beyond the scan window is omitted from the map, never set to `None`.** This is the one place the change can go quietly wrong. `None` means "looked up, no PR"; an absent key means "not fetched". A missed PR recorded as `None` reads to `stranded_items` (`funnel.py:2731`) as *"claim past its TTL with no PR"* — the scan's own blind spot rendered as a defect in the work, and it would grow as PR history grows. Omitting the key reuses the existing "not fetched" branch that `stranded_items` already handles via `pr_known`.
2. **Preserve the conflicting-PR verdict conditional.** `latest_verdict` is called only for an open PR whose `mergeable` is `CONFLICTING` (`funnel.py:3358`). Keep it, or the scan adds a verdict lookup per ticket and undoes the saving.
3. **Field fidelity is not a risk and needs no special handling.** `_ticket_pr` is already a `gh pr list` call differing only by `--head`, so lazily-computed fields (`mergeable` returns `UNKNOWN` on merged PRs) behave identically. This is a re-indexing, not a new data source.

Accept: fixture tests that (a) a ticket outside the scan window does **not** appear in `stranded_items` output as "claim past its TTL with no PR", (b) the map still distinguishes absent-key from `None`, and (c) the verdict lookup fires only for open CONFLICTING PRs. Plus the existing suite green.

Risk: standard

nateprich:
Closes #297.

## What

`ticket_pr_facts` built its map by calling `_ticket_pr` once per open ticket. That was 68 `gh pr list --head` requests on the 2026-09-08 board, measured with `rateLimit { cost }` at **93 of a full brief's 110 points (85%)** — the single largest GraphQL consumer in the system, and one that grows with the board.

It now builds a head-branch index from **one bounded `gh pr list --state all` per member repository**, adopting the shape `merged_pr_facts` already uses (`MERGED_PR_SCAN_LIMIT` with the `+1` truncation probe). Same signature, same return type, same map contract.

## The three things this could get wrong, each with a test

1. **Beyond-window tickets are omitted, never `None`.** A truncated scan cannot distinguish "no PR" from "PR older than the window". Recording `None` would reach `stranded_items` as *"claim past its TTL with no PR"* — the scan's blind spot rendered as a defect in the work, growing as PR history grows. Omitting reuses the existing "not fetched" branch that `stranded_items` reads via `pr_known`.
2. **The conflicting-PR verdict lookup stays conditional.** `latest_verdict` fires only for an open `CONFLICTING` PR. Unconditional, it would add a request per ticket and undo the saving.
3. **Field fidelity.** `_ticket_pr` was already a `gh pr list` differing only by `--head`, so lazily-computed fields behave identically. This is a re-indexing, not a new data source.

## Verification

`tests/test_ticket_pr_scan.py`, five tests:

- call count does not grow with ticket count (2 tickets and 30 tickets both cost one call)
- a beyond-window ticket is absent from the map, not `None`
- a genuinely missing PR inside a complete scan **is** `None`
- the verdict lookup fires only for the open `CONFLICTING` PR, not for mergeable or merged ones
- an unreadable response raises rather than returning a partial map

Full suite: **513 passed**. One unrelated pre-existing failure, `test_automation_drift.py` — 4 Codex automations are stale against `routines/codex-work.md`; that is #178's work and untouched here.

## Not done here

The measured before/after belongs to #299, which is blocked on #292/#295 landing so the number is read from `rateLimit { cost }` rather than projected. #271 is the worked example of why: an unmeasured cost model survived to a merged-intent PR. The ~18-point figure in the plan stays a projection until #299 measures it.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "e5497ecc8f4ea34a995e24e63bc0366fc4b66ba6",
  "reviewed_at": "2026-09-09T06:09:21.971268+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T06:09:24.124723+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #308, opened 2026-09-09
Expose GraphQL rate-limit data

### LINKED ISSUE #292: Add rateLimit { cost remaining resetAt } to GraphQL queries and surface it from gh_graphql

nateprich:
Closes #292.

## What changed
- add `rateLimit { cost remaining resetAt }` to every read query sent through `gh_graphql`
- expose the response block beside existing `data` roots, preserving callers such as `data["user"]`
- cover query shape, the inline option lookup, and response compatibility

## Verification
- `python3 -m pytest -q tests/test_graphql.py` — 3 passed
- `python3 -m pytest -q` — 511 passed; 1 pre-existing environment-only automation-drift failure for four installed schedules
- `git diff --check` — passed
- live GraphQL schema check for `rateLimit` — passed

Deliberately did not add `rateLimit` to mutations because GitHub exposes it on the query root.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "a41b543b358785e61b8c0b8e4791943aa266929e",
  "reviewed_at": "2026-09-09T04:28:27.090621+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-09T04:28:29.180141+00:00",
  "run": "d8701028e9b1",
  "voice": "agent"
}
```

Issue #239, opened 2026-09-08
Three different `core` rate-limit counters answer for the same token, so `gh api rate_limit` reports headroom a route does not have


nateprich:
# Measure the GitHub rate-limit bucket partition behind the contradictory counters, and write the result into LEARNINGS.md labelled measured

On 2026-09-08 three calls seconds apart on the same gh credential reported three different core counters with the same 5,000 limit and different resets, and on 2026-09-09 the /rate_limit endpoint reported full headroom on all fifteen resources while a real GraphQL call on the same token returned Remaining 0, Used 5012, and a reset nearly an hour earlier than the endpoint claimed. Any budget gate built on the endpoint reads a counter that is not the one being spent. The split-principal explanation is inferred, not established, and must not be written down as fact until someone measures it.

## What this does

One measurement pass, then one LEARNINGS.md entry. Hit four route families with the same credential, minutes apart, and record each response’s own headers — X-RateLimit-Resource, Limit, Used, Remaining, Reset — plus wall-clock time:

- an ordinary REST route (repos/...)
- the dependencies/* route that reported exhausted while others showed headroom
- a GraphQL call
- the GET /rate_limit endpoint itself, for comparison

Then determine how the buckets actually partition — how many counters there are, which routes map to which, and whether the windows imply different principals — and write the result into LEARNINGS.md labelled measured, since it contradicts what the documentation implies. The entry states the constraint for budget-gate work plainly: read the X-RateLimit-* response headers of a call actually made, never the /rate_limit endpoint.

## Rejected

- Building a budget gate, or any headroom check, on GET /rate_limit. Measured wrong twice: full headroom on every resource while a route was fully exhausted, and a reset nearly an hour later than the response header on the same token. Wrong in the dangerous direction, reporting headroom that does not exist.
- Writing the split-principal explanation into LEARNINGS.md as fact. It is the leading hypothesis from the 2026-09-08 windows, but the issue body says explicitly it was not established — recording it unmeasured would launder inference into precedent.
- Changing authentication or token setup to “fix” the split. There is no evidence the split is a misconfiguration, and touching credentials to chase an unmeasured hypothesis risks the working setup for nothing.
- Fixing #237’s doctor surface inside this plan. #237 owns that surface; this plan supplies the measurement and the constraint it must respect, and goes no further.

## Decided from precedent

[...]

PR #310, opened 2026-09-09
Record the REST-vs-GraphQL rate-limit divergence in LEARNINGS

### LINKED ISSUE #294: Record the REST-vs-GraphQL rate-limit counter divergence in LEARNINGS.md
Parent: #273.

What: two entries in `LEARNINGS.md`, labelled honestly per `AGENTS.md`.

1. **`measured`** — the REST `/rate_limit` endpoint and the real GraphQL counter are different counters, not a lagging copy. Quote both readings. Taken 2026-09-08 outside any outage, with budget to spare:

```
REST /rate_limit  ->  graphql: {limit: 5000, remaining: 5000, used: 0}
GraphQL rateLimit ->  {limit: 5000, remaining: 3174, used: 1826, resetAt: 03:52:06Z}
```

   A second instance was measured 2026-09-09 03:45Z and recorded on #239: `/rate_limit` reported `graphql: 5000/5000` across all fifteen resources while the same token's next GraphQL call returned `X-Ratelimit-Remaining: 0`, `X-Ratelimit-Used: 5012`, and the two disagreed on reset time by an interval. Include it — one reading is an anomaly, two is a property.

   The consequence worth stating: **`funnel doctor` can never source this from REST at all**, not even as a cheap pre-check.

2. **`documented`, not `measured`** — that `rateLimit` is itself a free field. Three consecutive `rateLimit`-only queries showed `used` flat across the first two and up by one on the third. On a shared token that is consistent with the field being free and the increment belonging to another consumer, but it does not establish it. Do not upgrade the wording without an isolated token to prove it on.

Accept: both entries present, newest first, each carrying its confidence label and its quoted evidence.

Risk: standard

nateprich:
Closes #294.

## What

Two entries in `LEARNINGS.md`, each labelled per the file's own convention.

**`measured`** — REST `/rate_limit`'s `graphql` resource is a *different counter* from GraphQL's own `rateLimit`, not a lagging copy. Measured twice, in opposite directions, which is what makes it a property rather than an anomaly:

| when | REST said | GraphQL actually |
|---|---|---|
| 2026-09-08, outside any outage | `graphql: 5000/5000, used: 0` | `remaining: 3174, used: 1826` |
| 2026-09-09 03:45Z | `graphql: 5000/5000` (all 15 resources) | `Remaining: 0, Used: 5012` |

The second reading has `used` **above** `limit`, and the two sources disagree on reset time by an interval.

**`documented`, not `measured`** — that `rateLimit` is itself a free field. Three consecutive `rateLimit`-only queries showed `used` flat across the first two and up by one on the third. On a shared token that is consistent with the field being free and the increment belonging to another consumer, but it does not establish it. Upgrading the label needs an isolated token — a credential, outside the agent capability boundary — so it stays `documented`. That is the honest label, and `AGENTS.md` is explicit that mislabelling `inferred` as `measured` is how a wrong belief becomes permanent.

## Why this is not a duplicate of the entry below it

`LEARNINGS.md` already carries *"`gh api rate_limit` can report full headroom while a route is fully exhausted"* (#239), about three disagreeing `core` counters. This is placed above it and cross-references it, because it closes a trap that entry leaves open: its advice is to read the **failing call's own response headers** instead of the endpoint. That is right for a REST route — but for GraphQL, headers give you `remaining` without `cost`, and cost is the half you need to attribute spend on a shared token. Only the in-query `rateLimit { cost remaining resetAt }` gives both.

## Verification

Documentation only. Full suite: **508 passed**, unchanged from main. One unrelated pre-existing failure, `test_automation_drift.py` — 4 Codex automations stale against `routines/codex-work.md`, which is #178's work and untouched here.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "c7e5be29d2963edefdcfcf2330eb497eb0fe607f",
  "reviewed_at": "2026-09-09T05:04:30.210819+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T05:04:32.210895+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #298, opened 2026-09-09
Delete _ticket_pr once the scan replaces its last caller


nateprich:
Parent: #272.

Depends on #297 — the scan must have replaced the loop before the helper has no caller.

What: delete `_ticket_pr` (`funnel.py:3334`) rather than leaving a per-ticket lookup in the file for someone to reach for again. Confirm no remaining caller first, including tests; rewrite or delete any test that only exercised it.

Same reasoning as #241: an unused path with passing tests reads as live code and drifts silently.

Accept: no reference to `_ticket_pr` remains in `funnel.py` or `tests/`, and the suite is green.

Risk: standard

Issue #272, opened 2026-09-09
`ticket_pr_facts` runs one PR lookup per open ticket — 85% of every brief's GraphQL cost


nateprich:
Shaped 2026-09-08 from precedent. No grilling: every question this raised was already
answered in this repo, and the one genuinely open technical question was settled by
measurement rather than by asking him.

## The defect

`_ticket_pr` (`funnel.py:3334`) runs `gh pr list --repo R --state all --head ticket/<n>`
**once per open ticket**. `ticket_pr_facts` calls it in a loop — 68 separate requests on
the current board, one per open ticket, to read facts that all live in the same list.

Measured against the live Project on 2026-09-08 with `rateLimit { cost }` read from the
response: one full `funnel brief` = **110 points**, of which `ticket_pr_facts()` = **93
(85%)**. `load_items()` is ~12 and `member_repos()` is 5.

This is the single largest GraphQL consumer in the system, and it scales with the board.

## The thing

**One `gh pr list` per member repo, indexed by head branch — the same one-call scan
`merged_pr_facts` already uses.** `ticket_pr_facts` keeps its signature, its return type
and its map contract; only how it fills the map changes.

### Measured, not modelled

Re-measured 2026-09-08 against the live repo, with `rateLimit { cost }` in the response:

| Query | Cost |
|---|---|
| One 100-PR scan, all seven fields, `reviews` nested 20 deep | **1 point** |
| One per-ticket `--head` lookup | **1 point** |

So 68 lookups → 1. The brief goes from ~110 points to roughly 18, and the remainder is
`load_items` and `member_repos`.

**The issue's open question is answered: `reviews` does come back from a list query.** A
single `gh pr list --state all --json number,state,url,headRefName,headRefOid,mergeable,
mergedAt,reviews` was run against `nateprich-projects/command-center` on 2026-09-08 and
returned all seven fields populated. No second round trip, and nothing needs to be
reintroduced per-ticket.

**Field fidelity is not a risk, and the reason is worth recording.** `_ticket_pr` is
*already* a `gh pr list` call — it differs only by `--head`. The scan is the same query
shape against the same endpoint, so lazily-computed fields (`mergeable` returns `UNKNOWN`
on merged PRs today) behave identically before and after. This change is a re-indexing,
not a new data source.

## Decided from precedent

- **The technique is not a new invention; it is already in the file twice.**
  `merged_pr_facts` (`funnel.py:3581`, landed as #269) does the bounded one-call scan with
[...]

Issue #250, opened 2026-09-08
The WIP-limit decision record names #129 and #161 as unfixed risks to watch, and both shipped


nateprich:
# Refresh the WIP_LIMIT watch paragraph to the risks that are actually live

`funnel.py:118-130` is the decision record for `WIP_LIMIT = 4`. Its closing paragraph
names two live risks — the unreadable prose dependencies of #129, and `merge_blockers`
not reading the `mergeable` it fetches (#161) — and tells a future reader that if the
raise goes badly, *"#129's #153/#154 and #161 are the fixes, not a lower number."*

**Both shipped and were accepted on 2026-09-08.** #129, #161, and also #175 and #177,
which the same paragraph leans on. The mergeable check is live on `origin/main`. So the
paragraph states as current two risks that no longer exist and points at fixes that
already landed.

It already cost something: reading line 127 against the code at HEAD during #236's
shaping made an accepted fix look unshipped, and resolving it took a detour.

## What this changes, and what it deliberately does not

**Scope: the facts, not the number.** `WIP_LIMIT` stays 4. Rewriting a decision record
in the course of a documentation fix is how a limit gets changed by someone who was not
making the decision — the issue raised this hazard itself, and it stands.

Rewrite the "what this does not fix, and what to watch" paragraph to say:

1. **#129 and #161 shipped**, with #175 and #177, on 2026-09-08. The queue no longer
   hands chained work to extra sessions with no memory of a decline, and the merge gate
   no longer passes a conflicting branch. Name them as *resolved*, keeping the numbers so
   the reasoning stays traceable.
2. **The risk that replaced them**, supplied by the #178 investigation on 2026-09-08:
   parallelism still multiplies whatever the queue hands out, but the multiplied thing is
   now a **stale prompt, not a stale queue**. All five Codex automations have drifted from
   `routines/codex-work.md`; four escalated-lane copies predate #268's decline handling,
   and the standard lane calls a `funnel next --not` flag that does not exist. At four
   concurrent sessions that is four runs executing an unparseable command instead of one.
   The fix is #178, not a lower number.
3. **The open question, stated as open**: whether 4 is still right now that the condition
   it was raised under has been met a second time. The paragraph should say this is Nate's
   and unanswered, rather than implying the number has been re-validated.

One paragraph in one file. No behaviour change, no test change.

## Rejected

- **Re-examining `WIP_LIMIT` in the same plan.** Your 2026-09-06 condition — *"the cap can
[...]

PR #311, opened 2026-09-09
Delete _ticket_pr and give cmd_show the shared scan

nateprich:
Closes #298. **Stacked on #307 (ticket/297)** — review that first; this PR's base is `ticket/297`, not `main`.

## The ticket's premise was wrong, and checking it found a second N+1

#298 assumed #297's scan left `_ticket_pr` with no caller. It did not. `cmd_show` still called it **once per child ticket**, so displaying a project with twenty tickets cost twenty requests — the same shape #272 exists to remove, in a second place nobody had counted.

So this does what the ticket actually asked (*"Confirm no remaining caller first"*) rather than deleting a function that was still live.

## What

- Extracts the bounded scan into **`ticket_pr_index(repo)`** — one `gh pr list --state all` per repository, returning `(index, truncated)`.
- `ticket_pr_facts` and `cmd_show` now share it.
- `_ticket_pr` is deleted.

`truncated` is returned rather than handled inside, because the two callers need **different** safe answers: `ticket_pr_facts` must omit a beyond-window ticket (a `None` there reaches `stranded_items` as a false "claim past its TTL with no PR"), while `cmd_show` is only printing and can show nothing. Burying that choice in the helper would have made one of them silently wrong.

## Also fixed

`awaiting_review`'s docstring named `_ticket_pr` as the handle it shares. Left alone, that is a comment pointing at a function that no longer exists — the exact defect family #250 is open about, created by this very PR. Repointed to `ticket_pr_index`.

## Verification

Three tests added to `tests/test_ticket_pr_scan.py`:

- `_ticket_pr` is gone from the module, so the next person cannot reach for it
- `ticket_pr_index` is the shared entry point
- a truncated scan reports `truncated=True` and caps the index at `MERGED_PR_SCAN_LIMIT`

Full suite: **516 passed**. One unrelated pre-existing failure, `test_automation_drift.py` — 4 Codex automations stale against `routines/codex-work.md`, which is #178's work and untouched here.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "efd4fd7d1a0edc0867d699365e2d561321170180",
  "reviewed_at": "2026-09-09T04:56:17.811085+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T04:56:19.177156+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #289, opened 2026-09-09
funnel doctor reports API calls per command against measured remaining budget


nateprich:
Parent: #237.

Depends on #292 and #295 — the `rateLimit` block and its per-run logging are the source this must read.

**Corrected 2026-09-09.** This ticket was first written to read the `X-Ratelimit-*` response headers of a live call. That is superseded by #273, which is `Broken`, was shaped first, and settles the mechanism: the measurement comes from `rateLimit { cost remaining resetAt }` **inside the GraphQL response**, because that yields a true per-query cost rather than only a remaining figure. Reading headers would have given headroom without cost, which is half of what "calls per command" needs.

What: `funnel doctor` reports how many API calls each command makes and what that costs, against the budget actually remaining — so headroom is visible before it runs out rather than after. Source every number from #292's `rateLimit` block.

**Never source this from REST `/rate_limit`.** It is not a lagging copy of the real counter, it is a different counter. Measured 2026-09-08 outside any outage: REST reported `graphql: {remaining: 5000, used: 0}` while the true GraphQL counter read `{remaining: 3174, used: 1826}`. Measured again 2026-09-09 03:45Z (#239): REST reported `graphql: 5000/5000` while the same token's next call returned `Remaining: 0, Used: 5012`. A doctor built on REST would confidently report full headroom while the next call fails — the exact failure this ticket exists to prevent.

Accept: `funnel doctor` prints calls-per-command with cost, and a remaining figure sourced from the GraphQL `rateLimit` block. A test asserts the REST endpoint is not consulted for this number.

Risk: standard

nateprich:
**Measured caveat for whoever builds this — the source this ticket names is incomplete.**

This ticket says to source calls-per-command from #292's `rateLimit { cost remaining resetAt }` block rather than from REST `/rate_limit`. The REST prohibition stands and is now measured twice (#294). **But the replacement has its own blind spot, measured 2026-09-09:**

`rateLimit { cost }` counts only queries issued through `gh_graphql`. `ticket_pr_facts`, `merged_pr_facts` and `awaiting_review` all reach GitHub through **`gh pr list`** — the `gh` CLI — and that spend is invisible to an in-query cost field while still landing on the same 5,000/hour GraphQL budget.

[...]

PR #312, opened 2026-09-09
Log GraphQL cost and remaining per run

### LINKED ISSUE #295: Log GraphQL cost and remaining per run
Parent: #273.

Depends on #292 — the `rateLimit` block must be surfaced before it can be logged.

What: log cost and remaining per run, so spend is observable rather than invisible until it hits zero. Today `gh_graphql` has no logging at all, so **the first symptom of exhaustion is total failure** — there is no earlier one. On 2026-09-08 `funnel.py` stopped working entirely (`brief`, `queue`, `next`, every gate command) for the better part of an hour, several times in one day.

This is also what turns #272's claimed drop from ~110 points to roughly 18 into a measured before/after rather than a projection.

Accept: a run's log carries per-command cost and remaining, and the totals for one `funnel brief` can be read back from it without instrumenting anything by hand.

Risk: standard

nateprich:
Closes #295.

## What

#308 made the `rateLimit` block available in each response; nothing accumulated it, so a run still could not say what it spent. This adds `graphql_spend()` and one stderr line per process.

**Cost is summed from each response's own `cost`, never from a `remaining` delta.** The token is shared — 1,826 points measured on 2026-09-08 belonged to something other than the measuring session — so a delta charges Nate's own `gh` use and every other agent to the funnel. A test pins it: two calls each reporting `cost: 2` while `remaining` falls by 50 must accumulate **4**, not 50.

**stderr, never stdout** — `begin`, `brief` and `next` emit JSON that routines parse. Verified `funnel brief | json.load` still parses. Printed from a `finally`, so a run that dies on an exhausted budget still reports what it spent; that is the run whose numbers matter most. A run that made no GraphQL call prints nothing rather than a row of zeros.

## A limitation I found by measuring, and did not paper over

**This counter does not see `gh` CLI spend.** `ticket_pr_facts` reaches GitHub through `gh pr list`, not `gh_graphql`, and that cost is invisible to an in-query `rateLimit { cost }` field while still landing on the same 5,000/hour budget.

Not theoretical. Measuring #297's before/after, this counter reported an identical **17 points for both branches**, while `remaining` deltas showed the true costs were **209 and 54** — a fourfold difference it could not see.

That limit is a property of the technique, not a bug to fix here: attributing a whole command needs a `remaining` delta, which is only trustworthy with the schedules paused. Recorded in `LEARNINGS.md` by #299 rather than left for the next person to trip over — and worth weighing when #289 builds `doctor`'s calls-per-command report on top of this.

## It caught the failure it exists for, while being written

The second consecutive `funnel brief` produced empty stdout and exited, budget gone:

```
graphql: 7 call(s), 17 point(s) spent, 130 remaining, resets 2026-09-09T04:52:19Z
```

Before this line existed, the only signal was a command that silently produced nothing.

## Verification

`tests/test_graphql_spend.py`, six tests: accumulation from the response, summing-not-differencing, a missing or malformed block not breaking the call, the report naming its numbers, silence when no call was made, and `unknown` said rather than guessed when `remaining` is absent.

Rebased onto main after #308 merged. Full suite: **524 passed**. One unrelated pre-existing failure, `test_automation_drift.py` — 4 stale Codex automations, which is #178's work.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "80c7eba2332a4a262787aafc0bc51c52f2efe7f2",
  "reviewed_at": "2026-09-09T05:56:50.709422+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T05:56:51.953741+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #313, opened 2026-09-09
Add the two-tier GraphQL reserve gate to funnel begin

nateprich:
Closes #296. **Stacked on #312 (ticket/295)**, and contains #293's commit because it needs that outcome to exist — merge #309 and #312 first.

## What

`funnel begin` declines when the GraphQL budget is below its tier's floor, returning the existing `{gate, do, why}` JSON rather than crashing. Coast to a stop rather than seize.

## Two tiers: reviewers keep going

Nate's call, 2026-09-08. New work consumes budget and adds in-flight PRs; review drains the lane and is cheap. One floor for everything risks a stall where approved PRs sit unmerged until the hour resets, and the undrained backlog makes the next hour worse than the one that triggered it — the same instinct as `plan.md`'s *"once a project is Building, its remaining tickets finish first"*.

**Accepted knowingly:** reviewers can still drive the budget to actual zero, so the last review runs of a bad hour may crash rather than decline cleanly. Two tiers make that rarer; they do not abolish it. Please don't "fix" it by collapsing the tiers.

## Three decisions worth reading

**Floors are loads-remaining, computed from the cost measured this run — not fixed point counts.** #272 changes per-load cost by roughly a factor of six, and a hardcoded number would rot silently. A test pins it: the same 1,500 remaining declines at a 100-point load and proceeds at a 12-point one.

**The tier is keyed on what the run would do, not on which agent asked.** So the split is consumers versus drainers, with no list of agent names to keep in step. That paid off immediately: `shape` landed on main from #306 while this was being written, and it slots onto the engineering floor by being a consumer, not by anyone remembering to add it.

**"No GraphQL call" is not "unreadable budget".** The first version conflated them and broke two existing `begin` tests by refusing runs that never asked. An unreadable block still fails closed, per `AGENTS.md`; no call at all proceeds, because refusing there is refusing on the absence of a question rather than the absence of an answer. In a real run it cannot arise — `load_items` queries before `begin` is dispatched.

## Rejected

**Backoff and retry.** Retrying against an exhausted hourly budget waits out a reset up to fifty minutes away, so a routine either sleeps through its own schedule or fails anyway — and spends credits sitting still. Declining cleanly *is* the backoff; the schedule supplies it.

## Verification

[...]

PR #309, opened 2026-09-09
Add a skipped-api-reserve heartbeat outcome

nateprich:
Closes #293.

## What

One new outcome in `heartbeat.py`'s `OUTCOMES`: `skipped-api-reserve`, for a run that declines because the GraphQL budget is below the reserve floor.

## Why it is its own ticket, and why it lands first

A budget decline is a **refusal, not a failure**. `heartbeat.py` already separates honest declines (`skipped-over-pace`, `skipped-usage-unknown`, `skipped-locked`, `skipped-blocked`) from `errored`, which means "tried and failed". Recording a reserve decline as `errored` would make the watchdog alarm on the system working correctly — and three `errored` runs in a week is its alarm threshold, so a bad hour would page.

The reserve gate (#296) needs this outcome to exist before it can record one, which is why this is separate and lands first.

## The design point worth noting

`.github/scripts/watchdog.py` ignores outcomes by **prefix** — its own docstring: *"Deliberately not reported: any `skipped-*` outcome and `nothing-to-do`."* So the name is what carries the classification, and a name outside that family would silently alarm. There is a test pinning that, rather than leaving it to the next person to notice.

Deliberately named distinctly from `skipped-over-pace`: that is the *token* budget gate in `usage.py`, this is the GitHub API reserve. Two different pools, two different refusals, and a shared name would make the heartbeat unable to tell you which one stopped a run.

## Verification

- `test_finish_accepts_skipped_api_reserve` — `heartbeat finish` accepts it and records it verbatim
- `test_the_reserve_decline_is_in_the_skipped_family` — pins the prefix the watchdog keys on
- `test_an_api_reserve_decline_is_not_an_alarm` — `watchdog.assess` returns healthy

Full suite: **511 passed** (508 on main, plus the 3 added here). One unrelated pre-existing failure, `test_automation_drift.py` — 4 Codex automations stale against `routines/codex-work.md`, which is #178's work and untouched here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #306, opened 2026-09-09
Offer tiered shaping work from funnel begin

nateprich:
## Summary

- Extend `funnel begin` with a final shaping job after review and optional breakdown work.
- Preserve `ideas()` ordering, filter by `required_tier()`, and require `usage.shaping_allowed()` headroom.
- Return one shapeable idea only; review and breakdown precedence remain unchanged.

## Deliberately not changed

- Routine prompt text is left for #136.
- The `routines/zcode.md:13` wording is left for #137.
- No Status, Class, routing vocabulary, or routine files were changed.

## Verification

- Focused tests: 15 passed (`test_begin`, `test_shaping_gate`, `test_guardrails`).
- Full local suite: 511 passed; one known pre-existing `test_automation_drift` failure for four installed schedules outside this ticket.
- Read-only AST syntax check and `git diff --check` passed.
- GitHub Actions pytest passed.

## Continuation

A concurrent run had already pushed `ticket/135` and opened this PR; this run preserved that implementation rather than force-pushing a duplicate. The exact remote commit and scoped diff were independently verified.

Issue #293, opened 2026-09-09
Add a skipped-* heartbeat outcome for a budget decline


nateprich:
Parent: #273.

What: add one new outcome to `heartbeat.py`'s `OUTCOMES` (`heartbeat.py:62`) for a run that declines because the GraphQL reserve floor was hit.

Why it is its own ticket and lands early: `heartbeat.py:62` already separates honest declines (`skipped-over-pace`, `skipped-usage-unknown`, `skipped-locked`, `skipped-blocked`) from `errored`, which means "tried and failed". A budget decline is a refusal, not a failure — recording it as `errored` would make the watchdog alarm on the system working correctly. The reserve gate needs this outcome to exist before it can record one.

Accept: the new outcome is accepted by `heartbeat.py finish`, is grouped with the other `skipped-*` outcomes wherever they are classified, and does not trip the watchdog. Fixture test for the classification.

Risk: standard

PR #322, opened 2026-09-09
Add the two-tier GraphQL reserve gate to funnel begin

### LINKED ISSUE #296: Two-tier GraphQL reserve gate in funnel begin, failing closed on an unreadable block
Parent: #273.

Depends on #292 (the `rateLimit` block must be readable) and #293 (the decline outcome must exist before a run can record one).

What: `funnel begin` declines to open a run when `remaining` is below a floor, returning the existing `{gate, do, why}` decline JSON (`funnel.py:3403`) instead of crashing. Coast to a stop rather than seize.

**Two tiers, decided by Nate 2026-09-08: reviewers keep going.** Engineering runs — Codex opening new work — stop at the higher floor; review and merge runs continue on a lower one. His reasoning: new work consumes budget and adds in-flight PRs, while review drains the lane and is cheap. One floor for everything risks a stall where approved PRs sit unmerged until the hour resets, and the undrained backlog makes the next hour worse than the one that triggered it. Same instinct as `plan.md`'s *"once a project is Building, its remaining tickets finish first"*.

**Accepted cost, knowingly:** reviewers can still drive the budget to actual zero, so the last review runs of a bad hour may crash rather than decline cleanly. Two tiers make that rarer; they do not abolish it. Do not "fix" this by collapsing the tiers.

Floors are **computed from live measured cost, not hardcoded points**: engineering 20 x the most recently measured load cost, review 5 x. Expressed in loads-remaining so they stay correct as per-load cost changes — which #272 is about to change by roughly a factor of six.

**Fail closed on an unreadable `rateLimit` block**, matching the existing `skipped-usage-unknown` branch — `AGENTS.md`: *"Missing usage data fails closed."*

The reserve check goes **after** the heartbeat start, alongside the existing gates: `cmd_begin`'s docstring is explicit that a run which cannot be recorded reads to the watchdog as never having happened, so the record comes before the decision.

Accept: fixture tests for three cases — engineering declines above the review floor while review proceeds; both decline below the review floor; an unreadable block fails closed. Each records the #293 outcome, not `errored`.

Risk: standard

nateprich:
Closes #296. **Replaces #313**, which was auto-closed when its base branch `ticket/295` was deleted on merge.

## Addressing the rejection on #313

zcode rejected #313 on sequencing, not content: it was stacked on unmerged #312 and **vendored #293's heartbeat commit**, so landing it would have put #293's change on main without its own review and stranded #309.

That is now resolved the way the verdict prescribed — *"wait for #309 and #312 to land, rebase ticket/296 onto main dropping the vendored #293 commit, re-offer"*:

- #309 merged 06:04Z, #312 merged 05:57Z
- rebased onto `main`, three commits, **no vendored commit**
- `git diff --stat main...ticket/296` touches only `funnel.py` and `tests/test_reserve_gate.py` — `heartbeat.py` is untouched, because `skipped-api-reserve` now comes from main

## What

`funnel begin` declines when the GraphQL budget is below its tier's floor, returning the existing `{gate, do, why}` JSON rather than crashing. Coast to a stop rather than seize.

**Two tiers, Nate's call 2026-09-08: reviewers keep going.** New work consumes budget and adds in-flight PRs; review drains the lane and is cheap. One floor for everything risks a stall where approved PRs sit unmerged until the hour resets, and the undrained backlog makes the next hour worse than the one that triggered it — the same instinct as `plan.md`'s *"once a project is Building, its remaining tickets finish first"*.

**Accepted knowingly:** reviewers can still reach actual zero, so the last review runs of a bad hour may crash rather than decline cleanly. Two tiers make that rarer, not impossible. Please don't collapse them to "fix" it.

## Three decisions

**Floors are loads-remaining, computed from the cost measured this run.** #297 has now landed and cut per-brief cost from ~209 to ~54 points; a hardcoded number would already be wrong. A test pins it: 1,500 remaining declines at a 100-point load and proceeds at a 12-point one.

**The tier is keyed on what the run would do, not which agent asked** — consumers versus drainers, with no agent list to keep in step. That paid off: `shape` arrived from #306 mid-flight and slots onto the engineering floor by being a consumer.

**"No GraphQL call" is not "unreadable budget".** The first version conflated them and broke two existing `begin` tests by refusing runs that never asked. An unreadable block still fails closed per `AGENTS.md`; no call at all proceeds.

## Verification

`tests/test_reserve_gate.py`, nine tests. Full suite: **544 passed**. One unrelated pre-existing failure, `test_automation_drift.py` — 4 stale Codex automations, tracked by #178.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "6331993f7f85fc1f276b533219059b16d5ada4bf",
  "reviewed_at": "2026-09-09T06:19:15.174115+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T06:19:17.165891+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #315, opened 2026-09-09
Human step: install the run clone and keeper job


nateprich:
Part of #171; discovered while implementing #204.

Human step: physical access to a machine

Action Nate must perform: On the Mac mini, clone nateprich-projects/command-center to ~/.claude/command-center-run, copy launchd/com.nateprich.command-center-run-keeper.plist from ticket/204 to ~/Library/LaunchAgents, and bootstrap com.nateprich.command-center-run-keeper after the engineering PR is merged.

Risk: standard

nateprich:
**Worked 2026-09-09. Sequencing corrected; the stated reason for this being a human step is wrong.**

**This and #204 were blocked on each other.** #204 carried *"Blocked on #315: Complete the human step before resuming."* This ticket says to bootstrap *"after the engineering PR is merged"* — #204's. Neither could move. The deadlock was invisible in `/funnel`: this issue is not `stranded` and #204 was `blocked`, so neither reached a gate. It surfaced only because Nate read the human-steps list and picked this off it by hand — which is the argument for that list existing.

Resolved in this ticket's favour: the sequence here was right. #204's block is cleared and its work is now open as **PR #350**, which had never been opened at all.

**`physical access to a machine` is not the real reason.** The Claude Code session doing this work runs **on the Mac mini** — `hostname` `mac-mini`, `hw.model` `Mac16,10`. Cloning the repo to `~/.claude/command-center-run` and copying a plist into `~/Library/LaunchAgents` are ordinary filesystem operations available to any agent running there.

**What actually gates this is authorisation, not capability.** `launchctl bootstrap` starts a persistent job that runs every five minutes indefinitely. That needs Nate's explicit say-so whether or not an agent can type it. A human-step marker is still correct here — the `reason` is what is wrong, and a wrong reason is worse than a vague one, because it sends the next reader looking for someone to physically walk to a machine that the agent is already sitting on.

Worth generalising when #89's family of work is revisited: `physical access to a machine` should be reserved for a machine the agent is *not* running on. Where it is, the honest marker is that the action changes persistent state and needs authorisation.

**Remaining sequence:** PR #350 reviewed and merged, then clone, copy the plist from `main`, then bootstrap on Nate's go-ahead.



```json
{
  "agent": "codex",
  "at": "2026-09-09T08:27:33.921711+00:00",
  "run": "561b3115d16e",
  "voice": "nate-relayed"
}
```

nateprich:
[...]

Issue #270, opened 2026-09-09
Human step: provision TickTick API token


nateprich:
Part of #176; discovered while implementing #207.

Human step: entering a credential

Action Nate must perform: Obtain a TickTick Open API token from the TickTick account/API-token UI and add it as the repository GitHub Actions secret TICKTICK_TOKEN so the watchdog can create alert tasks.

Risk: standard

nateprich:
**Blocked on #141:** this ticket is jamming the entire Codex lane, and the block is a temporary lever to unjam it, not a statement that the work is unwanted.

**What was happening.** Codex picked this ticket up on every run, discovered it is a human step it cannot perform, stopped, and recorded `errored` — producing nothing. Measured 2026-09-08: **25 errored runs in six hours**, continuous for the 96 minutes before this comment, with the same note each time: *"stopped: #270 is the existing human-step ticket; Nate must obtain a TickTick Open API token and add GitHub Actions secret TICKTICK_TOKEN; no repository changes or PR opened."* Meanwhile zcode and muse both reported `nothing-to-do`, correctly — there were no PRs to review, because Codex was not opening any.

**Why the deadlock.** #141 (`startable()` excludes human-step tickets) is the fix, and it sat at queue position 3 — directly behind this ticket at position 2. Codex could not reach the fix for the thing that was stopping it. The marker on this issue is already correct; #138 landed the marker and parser, but nothing reads it in `startable()` until #141 lands.

**The condition is #141, deliberately.** Once it lands, human-step tickets are excluded by their marker and this label becomes redundant — so #141 is genuinely the condition for *removing the label*, and naming it keeps this out of the "Unblock or park?" nag that an unparseable condition would produce (#189, #25).

**The real requirement is unchanged and is Nate's:** obtain a TickTick Open API token and add it as the repository Actions secret `TICKTICK_TOKEN`. No agent can do that — it is credential entry through an account UI, outside the capability boundary. Blocking here does not defer that; it only stops this ticket from burning every Codex run while it waits.

Nate decided this in a /funnel session on 2026-09-08 after I showed him the heartbeat evidence. He decided; I typed.



```json
{
  "agent": null,
  "at": "2026-09-09T02:13:22.999362+00:00",
  "run": null,
  "voice": "nate-relayed"
}
```

nateprich:
**Scope correction from Nate, 2026-09-08: the TickTick token already exists on the machine.**

[...]

Issue #147, opened 2026-09-08
funnel brief gains human-steps list and closed-with-vocabulary flag


nateprich:
Part of #89 (its ticket 10) — read #89 first. Depends on #138 (the marker) — do not start until it lands.

## What to build

Two additions to funnel brief, settled from precedent in #89:

- Open human-step tickets get their own list, the way parked already has one. They are work Nate owes, not a decision he owes, so they stay out of awaiting_decision() — the same distinction #85 restores for blocked items.
- The detective backstop for when every preventive layer fails: a project closing with access vocabulary in its plan and no human-step tickets is flagged. That is the case that otherwise leaves no trace at all.

## How to tell it worked

Fixture tests in tests/test_brief.py: an open human-step ticket appears in the new list with its reason and never in the decision items; a project closed with access vocabulary and no human tickets raises the flag; a project with neither stays quiet; parked and existing keys unchanged.

## Scope

funnel brief plus the two helpers and tests. Touches funnel brief, as do #58 and #81 — whichever lands second rebases. The contradiction scan itself is ticket 6 (#143); this ticket only surfaces its inputs at close time.

Risk: standard

Issue #141, opened 2026-09-08
startable() excludes human-step tickets


nateprich:
Part of #89 (its ticket 2) — read #89 first. Depends on #138 (the marker and parser) — do not start until it lands.

## What to build

startable() excludes marked tickets, alongside the blocked and awaiting-review exclusions it already makes. No agent can claim a human-step ticket, so no agent can satisfy one with a plausible diff — the failure #25 traced, where documentation describing the wiring would have passed every merge check.

## How to tell it worked

Tests: a marked ticket is never returned to any agent at any tier; an unmarked ticket under the same conditions is unaffected. The exclusion reads the marker through #138's parser — no second convention.

## Scope

The startable() exclusion and its tests only. The marker format is #138's; the brief lists are ticket 10's.

Risk: escalated — authorisation (this decides what agents may claim unattended)

nateprich:
Blocked on its explicit dependency #138. Verified against the repository on 2026-09-07: #138 is still open, HUMAN_STEP parsing is absent from main, and there is no remote ticket/138 branch or pull request to consume. Starting #141 now would either duplicate the marker convention or guess an interface the ticket explicitly requires it to reuse. Unblock when #138 lands on main; then remove the blocked label and implement only the startable() exclusion and its tests.



```json
{
  "agent": "codex",
  "at": "2026-09-08T06:34:46.240006+00:00",
  "run": "4172125ecba1",
  "voice": "agent"
}
```

nateprich:
**Blocked on #138:** `startable()` cannot exclude human-step tickets until the marker and its parser exist. Remove once #138 has landed.

_Restated in the convention #110 established. The prior block comment on this issue said the same thing in prose that `parse_block_comment` cannot read, so this item kept asking Nate "Unblock?" when its condition was already recorded. Nothing changed about the block itself._




```json
{
  "agent": "claude",
  "at": "2026-09-08T17:06:27.927632+00:00",
  "run": null,
  "voice": "agent"
}
```

nateprich:
Unblocked. Every condition this ticket named is now closed: #138 (closed 2026-09-08 16:35Z).

Nate decided this in a /funnel session on 2026-09-08, after I showed him each blocked ticket against the current state of its conditions. He approved lifting the label on #108, #141, #145 and #168 together. Executed by Claude on his instruction — he decided, I typed.

The `blocked` label answers one question only: is the prerequisite done. It is. Any risk bar this ticket carries is unaffected and still applies at review.

[...]

PR #339, opened 2026-09-09
Render the human-steps list /funnel already had the data for

### LINKED ISSUE #338: /funnel renders the human-steps list the brief already emits
Parent: #89.

What: teach `skills/funnel/SKILL.md` to surface `human_steps`, which `funnel brief` has emitted since #147 and which the skill reads past.

**The producer shipped and the consumer did not.** #147 added the list to the brief JSON; its scope was *"funnel brief plus the two helpers and tests"*. Nothing in #89's ten tickets touches the skill, so the field table documents fourteen keys without `human_steps`, and the render rule names only `needs_class`, `stale_locks_taken_over`, `stranded` and `in_motion` as things to surface when present. The data is correct and invisible.

**#141 made this sharper.** Before it landed, a human-step ticket kept announcing itself by being handed to an engineer run that wedged on it — noise, but noise Nate noticed. Since 2026-09-09 07:11 they are excluded from every agent queue and rendered by nothing he reads. Two are live right now: #270 (entering a credential) and #315 (physical access to a machine).

This is #89's stated outcome — *a ticket only Nate can do* must be distinguishable — so it belongs here rather than as a separate project.

**Scope: the skill file and a guard test.** Do not change `funnel brief`, `human_step_json`, or the marker; those are correct.

1. Add `human_steps` to the field table, saying what it is: work Nate owes, not a decision he owes — deliberately outside `total_needing_nate`, the same distinction #85 restores for blocked items.
2. Add it to the render rule. It should appear **whenever present**, with each item's reason, not folded into the decision list.
3. **A guard test so the next producer/consumer split is caught mechanically**: assert every top-level key `funnel brief` emits is named in `skills/funnel/SKILL.md`. This defect was a documented key going unread, and the repo prefers detection over an agent noticing (`AGENTS.md`, and #97).

Accept: the field table and render rule name `human_steps`; the guard test passes at HEAD and fails when a brief key is added without documenting it — show that in the PR by adding a throwaway key and quoting the failure.

Risk: standard

nateprich:
Closes #338.

## Reviewer note: this PR changes `skills/` on purpose

The routine says a ticket PR touching `skills/` fails review *unless its own ticket asked for the change*. **#338 asks for exactly this** — the skill file is the entire subject. No other guarded path is touched: `funnel.py` is unmodified, and `git diff --stat funnel.py` is empty.

## The defect

#147 added `human_steps` to `funnel brief`. Nothing taught `skills/funnel/SKILL.md` to read it — the field table documented fourteen keys without it, and the render rule named only `needs_class`, `stale_locks_taken_over`, `stranded` and `in_motion` as things to surface. The data was correct and invisible.

**#141 turned that from untidy into the only failure mode.** Before it, a human-step ticket kept announcing itself by being handed to an engineer run that wedged on it — noise, but noise Nate noticed. Since 2026-09-09 07:11 no agent can be handed one, so an unrendered human step is invisible everywhere. Two are live now: **#270** (entering a credential) and **#315** (physical access to a machine).

This is #89's stated outcome — *a ticket only Nate can do* must be distinguishable — which is why the ticket sits under #89 rather than standing alone. It also means #89 is no longer acceptable at 10/11, which is correct: accepting it while its outcome was undelivered is the #25 failure the breakdown skill warns about.

## Scope grew for a reason

The audit found **seven** undocumented keys, not one:

`awaiting_breakdown`, `closed_with_access_vocabulary`, `generated_at`, `human_steps`, `parked`, `rejected_merges`, `unattended_merges`

So the producer/consumer split is structural. Documenting only `human_steps` would have meant scoping the guard narrowly enough to pass, which is worth nothing. All six substantive keys are now in the field table, **described from what they actually contain** — I read each one's live value and its function docstring rather than inferring from the name.

`human_steps` gets its own line in the render rule: never folded into the decision list, never counted in `total_needing_nate`. It answers a different question — not *what do you have to decide* but *what is waiting on you to go and do*.

## The guard

`tests/test_funnel_skill_documents_brief.py` asserts every key `cmd_brief` emits is named in the skill. It parses the dict literal from source rather than running `funnel brief`, so it needs no network and no token. `generated_at` is the single named exception, as transport metadata — kept explicit so the exception list cannot be quietly widened to whatever is failing.

**Demonstrated failing**, per the ticket's acceptance criterion — adding a throwaway key to `cmd_brief`:

```
AssertionError: these keys are emitted by `funnel brief` and never mentioned in
skills/funnel/SKILL.md, so nothing renders them:
    throwaway_probe

Document each in the field table and say when to surface it, or add it to
NOT_RENDERED with a reason if it is transport metadata.
```

Reverted immediately; `funnel.py` is unchanged in this branch. A second test asserts the parser actually found keys, so the guard cannot pass vacuously if the source shape changes.

## Verification

Full suite: **551 passed**. One unrelated pre-existing failure, `test_automation_drift.py` — 4 stale Codex automations, tracked by #178.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "d6b8565487a09d20e2dae326f098d7ea3484e99d",
  "reviewed_at": "2026-09-09T07:32:24.976980+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T07:32:27.310118+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #171, opened 2026-09-08
Routines run `funnel.py` from Nate's working tree, which nothing keeps current, so every merged fix stays dormant until a human pulls


nateprich:
Shaped 2026-09-08. One concern: **the code the agents run is whatever a human last pulled.**

## The defect

All 22 `funnel.py` invocations across `routines/*.md` are the absolute path `/Users/nateprich/.claude/command-center/funnel.py` — **Nate's working tree**, not a clone of `main`.

Agents are explicitly forbidden from updating it. `routines/muse.md:133` and `routines/zcode.md:135`: *"never touch `/Users/nateprich/.claude/command-center`"*, citing a 2026-09-06 run that ran `git pull --ff-only` inside it and moved his checkout underneath him. Nothing else pulls it — grep for `git pull` across `routines/`, `scripts/` and `.github/` returns only those two prohibitions.

**So every merged fix is dormant until a human pulls.** Measured on 2026-09-07/08: the checkout was behind at 22:15 (PR #116 merged, yet `funnel capture --help` still offered `--needs-shaping`, which led to a wrong conclusion that #46 had not landed), and behind again at 23:40, and again at 08:24. Each time it advanced only because Claude pulled by hand in an interactive session.

**The consequence with teeth:** #55 makes `cmd_merge` auto-close a finished project — the change that removes Nate's manual accept ritual. It will merge and **not take effect**. The same is true of every fix the funnel ships for itself, #161 and #133 included.

**Two constraints, each right, jointly contradictory.** Agents must not move his tree under him. Routines need current code. Nothing today resolves that; it is resolved by a person remembering.

## Decided from precedent

- **A separate clone, and leave his working tree alone.** The prohibition is correct and must survive: he has uncommitted work in that tree, and a background `git pull` under an interactive session is a real hazard, not a theoretical one. Precedent for a sibling directory already exists — Muse runs against `~/.claude/command-center-muse-workspace`.
- **Kept current by its own scheduled job**, in the same shape as the existing launchd schedules, doing `git pull --ff-only` and nothing else. Fast-forward only: the clone must never acquire local state to reconcile.
- **Read-only in practice.** Nothing writes to it. Engineers already use their own clones; this exists solely so `funnel.py` is current when a routine invokes it.

## Needs you

Nothing, but see "Watch for" — the migration has a step that is easy to get wrong and it is worth reading before this is started.

## What to build

[...]

PR #350, opened 2026-09-09
Add command-center-run clone kept current by its own launchd job

### LINKED ISSUE #204: Add command-center-run clone kept current by its own launchd job
Parent: #171.

Build the execution source the routines will point at: a separate clone at the sibling path `~/.claude/command-center-run`, plus a launchd job that fast-forwards it on a schedule at least as often as the fastest routine fires (standard schedule fires every fifteen minutes, so the keeper must fire at least that often).

Scope:
- Clone the command-center repo to `~/.claude/command-center-run`. The job does `git pull --ff-only` and nothing else. Fast-forward only: the clone must never acquire local state to reconcile.
- Ship the keeper as a plist under `launchd/` (same shape as the existing `launchd/com.nateprich.command-center-*.plist` schedules), installed to `~/Library/LaunchAgents` and loaded. See `tests/test_launchd_drift.py` for the repo-copy-must-match-installed-copy contract.
- Read-only in practice: nothing writes to the clone. Engineers keep using their own clones; this exists solely so `funnel.py` is current when a routine invokes it.

Do NOT touch in this ticket:
- `routines/*.md` invocations (that is the follow-up ticket).
- `~/.claude/command-center` (Nate's tree) and the prohibition lines in `routines/muse.md` and `routines/zcode.md` ("never touch `/Users/nateprich/.claude/command-center`"). They stay exactly as written.
- `skills/funnel/SKILL.md` (deliberately keeps the old path for interactive sessions; see parent plan).

How to tell it worked:
- `git -C ~/.claude/command-center-run status` shows a clean clone of `main`; leaving Nate's tree stale or dirty has no effect on it.
- A commit merged to `main` appears in the clone after the next keeper fire with no human pulling (keeper log shows the ff-only pull).
- New plist has a repo copy under `launchd/` and `tests/test_launchd_drift.py` passes where the plist is installed (skip elsewhere is honest).

Risk: escalated — unattended pull job deciding what code every later agent run executes.

nateprich:
Closes #204.

Adds `launchd/com.nateprich.command-center-run-keeper.plist` — the job that keeps the read-only routine checkout at `~/.claude/command-center-run` current with a five-minute `git pull --ff-only`, so a merged fix reaches the routines without a human pulling (#171).

`tests/test_launchd_drift.py` gains the keeper to its parametrised set, so the installed copy is compared against the canonical one in the repo. It **skips** where the plist is not installed, which is what makes merging this before the machine-side install safe.

### Why this sat unmerged

#204 was blocked on #315 ("complete the human step before resuming"), while #315 says to bootstrap the job *"after the engineering PR is merged"* — a deadlock in which neither could move, and one invisible in `/funnel`: #315 is not `stranded` and #204 is `blocked`, so it never reaches a gate.

The block was backwards. It would have been justified if merging first broke CI, but the drift test skips when the plist is absent, exactly so it is "a real check on the Mac that runs the schedule and silent everywhere else." Nate cleared it on 2026-09-09.

### Branch state

`ticket/204` was 11 behind `main`. Merged `origin/main` in rather than rebasing — AGENTS.md:205 requires asking before rewriting history or force-pushing, and nothing here needed it. Clean merge, no conflicts.

### Test suite

**562 passed, 1 skipped, 1 failed.** The failure is `test_automation_drift.py::test_every_codex_automation_matches_the_routine`, and it is **pre-existing on `main`** — verified by running that file against a clean `origin/main` checkout, where it fails identically. It is not caused by this branch and this branch does not touch those files.

It is worth reading on its own terms though: 4 of 5 Codex automations no longer match `routines/codex-work.md`, which means those schedules are executing stale instructions right now. Reported separately rather than fixed here.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "b9d745ab4413e12a569ce7a9b4ac1d41852eed2a",
  "reviewed_at": "2026-09-09T08:37:11.687023+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T08:37:14.019919+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #229, opened 2026-09-08
Nightly watch job: run both checks on schedule, silent when clean


nateprich:
Parent: #162.

Depends on the pricing-watch and release-watch tickets above having landed — this job composes those two checks, so it cannot be verified until they exist. If either is still open when this ticket is offered, decline it.

Scope: nightly entry point that runs the pricing check and the release check independently and sends one TickTick task per change found. Silent when nothing changed. A fault reported by either check (could-not-check) is surfaced as a fault, never swallowed into silence.

Out of scope: the checks themselves — they live in the two sibling tickets. This ticket is the schedule plus the composition.

Verify with fixtures only, no network: a clean run produces zero notifications and zero writes; a run where a check reports changes produces one TickTick task per change; a run where a check reports could-not-check surfaces a fault.

Risk: standard

nateprich:
**Native dependency edge added, 2026-09-10: blocked by #227.** This ticket's body says it depends on the pricing-watch (#227, open) and release-watch (#228, closed) tickets, in prose that names no number, so the queue could not see it: since 08:00Z today `begin` offered this ticket to Codex six times, each run declined it for the same reason, and three of those runs never recorded a finish. The edge is what the breakdown skill requires and what #357's backfill could not infer here; `startable()` now withholds this until #227 closes. Written by Claude in session under Nate's standing direction to unwedge; no Status or Class change.



```json
{
  "agent": "claude",
  "at": "2026-09-10T13:34:07.954261+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #413, opened 2026-09-09
Add read-only model release watch

### LINKED ISSUE #228: Nightly release watch: notify on newer model version, change nothing
Parent: #162.

No dependency — startable any time. Independent of #165 (releases write nothing) and of the pricing-watch ticket.

Scope: for each model in heartbeat.MODEL_SOURCES, detect a newer version. On a new version, notify via TickTick (one task per new version) and stop. No configuration is written and no model string is changed anywhere.

Rejected alternative (fails this ticket if present): any automatic model switch, within a version family or outright. Nate declined both on 2026-09-08: a model swap changes cost, quality and behaviour at once and a quality regression would present as odd agent behaviour, not as an upgrade. The watch never switches a model.

Verify with fixtures only, no network, against captured vendor responses: a new model version produces a notification and zero configuration writes — assert the zero writes explicitly, it is the decision above. No new version stays silent. A response that cannot be parsed reports a fault (could-not-check), never silence.

Risk: standard

nateprich:
Closes #228

## Summary
- derive the models in use from heartbeat model sources
- parse captured provider model catalogues and compare only the same model family
- emit one TickTick task description per newer version and surface malformed or missing responses as `could-not-check` faults
- keep the release watch read-only: no model string or configuration writes

## Verification
- `python3 -m pytest tests/test_release_watch.py -q` — 11 passed
- `python3 -m pytest tests/ -q` — 676 passed

## Deliberate boundary
The check accepts captured vendor responses and exposes a notifier callback; it does not make live vendor or TickTick network calls, read credentials, or schedule the job. The live transport and nightly composition are the scope of #229. No automatic model switch is implemented.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "245cc028d4c26df65f458c3bccee6d0dc549886f",
  "reviewed_at": "2026-09-09T20:06:35.692674+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T20:06:38.242115+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #430, opened 2026-09-09
Guard funnel API calls against item-count scaling

### LINKED ISSUE #288: Guard test: a funnel load's API call count scales with pages, not items
Parent: #237.

What: add a test that counts `_gh_json` and `gh_graphql` invocations across a fixture load and asserts the count does **not** grow with item count. Parameterise one fixture over two sizes (e.g. 10 and 100 items) and assert the call counts are equal. Cover the three known per-item shapes: `load_items`, `_load_block_comment` (one `gh issue view` per open blocked item) and `_ticket_pr` (called per candidate).

The mechanism is already proven — the guard added in ff35327 asserts `calls == []` for the dependency path. This generalises it so the *shape* is caught at development time rather than by an outage. ff35327 was found by hand, after ~78 serial REST calls per command exhausted the budget and every routine failed closed.

Accept: the test passes at HEAD, and fails when a per-item call is reintroduced into any of the three paths — demonstrate that in the PR by temporarily reintroducing one and quoting the failure. A test that only passes on today's data does not count.

Risk: standard

nateprich:
Closes #288.

## What changed

Added `tests/test_api_scaling.py` with a fixture load containing 10 and 100 items, one Project page, one open blocked item, and one member repo. The test counts both `gh_graphql` and `_gh_json` calls and requires the counts to remain equal as the board grows. It exercises `load_items`, the `_load_block_comment` path, and `ticket_pr_facts`, which is the current replacement for the removed `_ticket_pr` lookup.

## Verification

- Focused regression set: 26 passed.
- Full suite: 696 passed.
- Proof of sensitivity: temporarily reintroduced the per-item dependency read in `load_items`; the guard failed with `10 items -> 11 JSON calls` versus `100 items -> 101`. The temporary production change was removed before commit.

## Deliberately not changed

This ticket is the regression guard, so no production code was changed. The intentional one-comment-read-per-open-blocked-item behavior remains; the fixture holds that count at one so the guard detects unwanted scaling with the total board instead of changing that separate behavior.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "12dedbe1d8ebce213dbca54a4fd70d71d8d9df97",
  "reviewed_at": "2026-09-09T21:19:18.011986+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-09T21:19:20.323265+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #482, opened 2026-09-10
Muse review bound tests time out on host


nateprich:
# Muse review bound tests time out on host

Proposed class: Improve

## Problem
Two tests in tests/test_muse_review_bound.py time out on the host: test_a_run_past_the_bound_is_killed_and_finished_errored and test_a_run_inside_the_bound_is_left_alone. Each exceeds the 30 second outer subprocess timeout while running scripts/muse-review. The rest of the suite passes (829 to 940 passed in repeated runs, with only these two plus occasional unrelated drift failures). Repeated observations across many runs record the same signature, so this is a stable host timing or test-harness defect, not a one-off slow run.

## Goal
Make both bound tests deterministic and fast on the host without changing production review-bound semantics: a run past the bound is still killed and finished as errored, and a run inside the bound is still left alone.

## Scope
In scope:
- Measure where the 30 seconds go in both cases on the host.
- Fix the test harness or test double around scripts/muse-review so both cases complete well under the outer timeout.
- Keep assertions on kill versus leave-alone behavior identical in meaning.

Out of scope:
- No change to production kill semantics, bound defaults, or review verdicts.
- No change to any other scheduled routine, gate, or permission surface.
- No repository creation or transfer.

## Plan
1. Reproduce and time both tests on the host with timestamps around scripts/muse-review startup, bound wait, and teardown, and record the breakdown.
2. Isolate the slow step (model or network wait inside scripts/muse-review versus harness wait logic) and choose the smallest deterministic fix: stub the review runner behind the existing script interface, shorten the exercised bound values, or scope a longer outer timeout to exactly these two tests.
3. Apply the fix, rerun the two tests plus the full file repeatedly, and report timings and pass counts.

## Rejected alternatives
- Deleting or permanently skipping the two tests: rejected because the kill versus leave-alone boundary is the behavior under test, and skipping would remove the only check on it.
- Raising the global pytest or subprocess timeout: rejected because it slows the whole suite and masks the underlying nondeterminism instead of removing it.
- Changing production bound or kill behavior to fit the tests: rejected because the tests exist to pin that behavior, so fitting production to the harness inverts the check.
- Treating this as a product outage: rejected because observations consistently describe a host-only harness timing defect associated with closed #392, with production behavior unaffected.

## Decided from precedent
- Class Improve stands, per the capture note on #482: this is an observed test-harness and runtime-timing defect, not a new product feature. No class change is proposed.
[...]

Issue #467, opened 2026-09-09
Codex automation drift remains on hourly schedule


nateprich:
Observed in the full suite: 760 tests passed, but tests/test_automation_drift.py::test_every_codex_automation_matches_the_routine failed because the live command-center-tickets-hourly automation no longer matches routines/codex-work.md. Class Improve: this is an operational sync/configuration defect, not evidence of a product runtime failure.

```json
{
  "agent": null,
  "at": "2026-09-09T21:24:21.046027+00:00",
  "run": null,
  "voice": "agent"
}
```

```json
{
  "agent": null,
  "at": "2026-09-09T21:24:21.046027+00:00",
  "run": null,
  "voice": "agent"
}
```

nateprich:
Class selected: Improve. The observed failure is a stale live automation copy; the repository sync/configuration path needs improvement, but no product runtime failure was observed.



```json
{
  "agent": "codex",
  "at": "2026-09-09T21:25:34.117805+00:00",
  "run": "37fa3c9f7e4d",
  "voice": "agent"
}
```

nateprich:
Observed again in this run's full suite: 765 tests passed, while tests/test_automation_drift.py::test_every_codex_automation_matches_the_routine failed because the live command-center-tickets-hourly automation does not match routines/codex-work.md. Class Improve: this is the already-tracked operational sync/configuration defect, not a new product runtime failure; no duplicate capture filed.



```json
{
  "agent": "codex",
  "at": "2026-09-09T22:46:29.020750+00:00",
  "run": "96053123944a",
  "voice": "agent"
}
```

nateprich:
Observed again in this run's full suite: 775 tests passed, while tests/test_automation_drift.py::test_every_codex_automation_matches_the_routine failed because all five live Codex automations do not match routines/codex-work.md. Class Improve: this is the already-tracked operational sync/configuration defect, not a new product runtime failure; no duplicate capture filed.



```json
{
  "agent": "codex",
  "at": "2026-09-09T22:58:00.710262+00:00",
  "run": "b5b1d1759dd2",
  "voice": "agent"
}
```

nateprich:
Observed again in this run's full suite: 775 tests passed, while tests/test_automation_drift.py::test_every_codex_automation_matches_the_routine failed because all five live Codex automations do not match routines/codex-work.md. Class Improve: this is the already-tracked operational sync/configuration defect, not a new product runtime failure; no duplicate capture filed.



```json
{
[...]

PR #486, opened 2026-09-10
Report per-command API usage in funnel doctor

### LINKED ISSUE #289: funnel doctor reports API calls per command against measured remaining budget

nateprich:
Closes #289.

## What changed

- Added one shared `gh` execution wrapper that counts every attempted call, separating GraphQL calls from `gh` CLI calls so CLI spend is visible without assigning it a guessed point cost.
- Kept GraphQL cost accumulation and latest remaining/reset values sourced from each response's `rateLimit` block.
- Added the `API usage` doctor check, which reports the current command's call count, measured GraphQL point cost, and measured remaining headroom.
- Reset the in-process measurement at CLI command start so repeated `main()` calls do not blend readings.
- Added focused tests proving the split accounting, response-sourced remaining value, unreadable-budget reporting, and no REST `/rate_limit` lookup.

## Verification

- `python3 -m pytest tests/test_graphql_spend.py tests/test_doctor.py` — 81 passed.
- `python3 -m pytest` — 845 passed; 3 pre-existing host-only failures: five protected Codex automation copies drift from the routine (#467), and both Muse review-bound subprocess tests exceed the local 30-second harness timeout (#482). This run recorded evidence on those existing issues; no duplicate defects were filed.
- `git diff --check` and AST parsing passed.

## Deliberately not done

- No REST `/rate_limit` query and no remaining-delta cost inference.
- CLI calls are reported separately because their responses do not expose the per-query GraphQL point cost.
- The pull request is not merged.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "38155fe375654ac44f916c94702da996419c9e7c",
  "reviewed_at": "2026-09-10T06:02:09.665549+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-10T06:02:13.998630+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #509, opened 2026-09-10
Every routine finishes with the run id its own begin printed, never one seen earlier in the session


nateprich:
Parent: #497.

Depends on #505 — the refusal that backs the wording must exist first.

What: in `routines/codex-work.md`, `routines/muse.md`, `routines/zcode.md` and `routines/CLAUDE.md`, the finish instruction says: pass the run id printed by **this run's** `begin` output, never an id from an earlier begin in the same session; a finish refused for a mismatch names the run to use — use it. Keep the existing guidance against wrapping the id in a command substitution. Then re-sync the Codex automations from the Mac (`scripts/sync_codex_automations.py`) — a Mac-side step for the merger, recorded in the PR.

Accept: `tests/test_routine_contents.py` asserts each live routine carries the this-run wording; the suite passes; the PR body records the sync.

Scope: the four routine files and tests. No `funnel.py` or `heartbeat.py` change.

Risk: standard

Issue #505, opened 2026-09-10
begin binds the work it issues to its start record, and finish refuses an outcome for a run that did not issue that work


nateprich:
Parent: #497.

What: `funnel begin` writes the work it issued — the claimed ticket, the review PR, the breakdown or shape ref — into the heartbeat start record (or a begin event carrying the run id), so the binding is durable. `heartbeat finish` with an explicit `--run` verifies the id is open **and** that any work named on the finish (`--merged`, the ticket in the note is not parsed — use the bound ref) is consistent with that run's binding; a mismatch is refused with a message naming the still-open run to use, and the refused outcome is filed as an event on the right run — never dropped, never filed under the wrong run. `resolve_run`'s exactly-one-open inference for an omitted id is unchanged. `append` never raises for bookkeeping; a verification failure degrades to a recorded event.

Accept (fixtures only, no network, spool in tmp_path): an explicit finish for an already-finished run is still refused (regression); an explicit finish naming work that a different open run issued is refused, names that run, and files the outcome as an event on it; a finish for the run that issued the work succeeds as today; an omitted run id with exactly one open start still resolves; two overlapping same-agent starts each finish under their own id.

Scope: `funnel.py` (begin), `heartbeat.py` (finish, append, read), tests. No routine text (a sibling ticket), no watchdog change.

Risk: standard

Issue #497, opened 2026-09-10
Heartbeat can record a different scheduled run's outcome


nateprich:
# Bind every finish to the run that began it

## Problem
Run 2593a91ac8f2 (codex, 2026-09-10) began, found the single-in-motion lock held by command-center#229, and finished nothing-to-do at ts 1789020318 with the note "nothing — lock held by nateprich-projects/command-center#229 (claimed about 1 hour ago); no ticket claimed or changed". The same run id later did ticket command-center#492 work and opened PR #496, and the explicit done finish was refused as already finished. That refusal saved this instance only by luck of ordering: had the stale id still been open, the done outcome would have been filed under the earlier tick run, while the run that actually did the work would have stood open until the watchdog called it dead. PR #496 merged at 06:26 UTC under muse run f11192b186b2 ("done merged PR #496", ts 1789021588), so the durable record for the #492 work now lives under a different run than the one that opened the PR.

Corroborating evidence, all read from heartbeat records without changing anything:
- codex records show start 2593a91ac8f2 at ts 1789020113 and finish 2593a91ac8f2 nothing-to-do at ts 1789020318.
- Same-agent starts interleave in normal operation: start de868ca483c2 at ts 1789016740, start 7f242ca8e6cf at ts 1789016856, finish 7f242ca8e6cf nothing-to-do, then finish de868ca483c2 done merged PR #484. Two schedules share one agent file, so overlapping starts are routine.
- codex finish 7ed613d1e286 nothing-to-do ("command-center#229 is locked ... claimed about 1 hour ago") shows the lock contention behind the first outcome.

## What changes
1. funnel begin binds the issued work to the start it just created: the claimed ticket, review, breakdown, or shape ref travels with the run id in the start record or a begin event, so a later finish checks consistency instead of trusting a bare id.
2. heartbeat finish with an explicit run id verifies the id is open AND bound to the begin of the run that is finishing. A mismatch is refused with a message naming the still-open run to use, and the orphaned outcome is filed as an event on the right run — never silently dropped, never filed under the wrong run.
3. A reconcile pass closes work-issuing starts whose ticket PR demonstrably merged under a different run: the merge record is attributed exactly once and the start is closed, instead of leaving it for the watchdog to report dead.
4. Routine text in all four routine files: finish with the run id printed by the begin output of this run, never an id seen earlier in the session. Regression tests cover the existing double-finish refusal and the new cross-filing refusal.

[...]

PR #515, opened 2026-09-10
Bind begin's issued work to its run, and make finish check the binding

nateprich:
Ticket #505 (project #497, `Broken`).

## What

Run 2593a91ac8f2 began, finished nothing-to-do, and the same id was reused for later ticket work; the explicit `done` finish was refused only because the stale id happened to be finished first. `funnel begin` now records a `bind` heartbeat record naming the ticket, PR, breakdown or shape it issued (both the engineer path and the reviewer path), and prints it under `bound`. `heartbeat finish --run` checks a named `--merged` PR or the new `--work` ref against that binding: a mismatch is refused with a message naming the still-open run to use, and the outcome is filed as an event on that run — or as an unattributed event carrying the open candidates when none is open — never dropped, never filed under the wrong run. Runs with no binding (a stop run, or a start from before bindings existed) are trusted; the exactly-one-open inference for an omitted id is unchanged; a binding that cannot be written is reported and never stops the run. Readers that count starts and finishes ignore the new phase.

## Verification

- `tests/test_run_binding.py`: the bind record and its reader; bind records are not starts or finishes; the run that was issued the work finishes as before; a finish naming another run's work is refused and filed on that run; `--work` is checked and an unattributed event carries candidates; a run with no binding is trusted; a second finish is still refused; overlapping same-agent starts each finish under their own id; an omitted id with one open start still resolves; `begin` binds the ticket it issues and a stop run binds nothing; a binding that cannot be written does not stop the run.
- `tests/test_begin.py`: the existing order-of-operations test passes unchanged; a module fixture stubs the binding writer so tests never touch the live spool (the first run without it did, and the records were removed by hand).
- Full suite locally: everything passes except `tests/test_muse_review_bound.py`, which times out on this loaded host (#482) and is unrelated.

Routine wording to pass `bound` back on `finish --work` is ticket #509, blocked on this.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "708c7e1cb8a05b9eb6aa812268c3e5fb724d2560",
  "reviewed_at": "2026-09-10T07:40:27.544527+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-10T07:40:29.801306+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #540, opened 2026-09-10
Reuse one funnel load across each Muse run

### LINKED ISSUE #291: Collapse the Muse run's five funnel invocations into one load
Parent: #274.

Depends on #152 and #272 — #152 is rewriting the exact surface this extends (`funnel begin` gaining a claiming ticket path with one-call opening), so building against today's `begin` would be building against code that is being replaced; and #272 removes ~85% of brief cost on its own, which may leave nothing here worth collapsing.

What: **measure first, then collapse.** Re-measure per-run funnel cost once both have landed, and read what #152 actually delivered — it may already cover the Codex lane, leaving only the Muse run's five calls (`begin`, `review`, `review`, `merge`, `brief`), which today re-page the same board five times in a couple of minutes with nothing changing in between. Then collapse those into a single-process path under one `load_items()`.

**No cache.** `AGENTS.md`: *"If you find yourself adding a cache, a `.json` of record, or a lock file, you have taken a wrong turn."* Its one sanctioned exception (the statusline cache) exists because those numbers are unreadable outside a live session; the Project is readable at any time, so the exception does not extend. A short-TTL cache is also intolerable to the claim lock, the one piece of state where a stale read makes two agents take the same ticket.

**The claim lock must keep reading GitHub at claim time.** A claim served from an in-process snapshot older than its own command is a departure from the plan.

Use the corrected numbers, not the capture's: `load_items()` ~12 points, a full `brief` 110, steady state ~2,300/hour against 5,000. Five invocations is ~175 points per run, not the ~1,800 first modelled.

Accept: **either** one process per Muse run, with measured points-per-run recorded before and after; **or**, if the re-measurement shows the multiplier is noise, a `LEARNINGS.md` entry carrying that measurement and this ticket closed without the collapse. Recording "not worth doing" with evidence is a correct outcome here, not a failure.

Risk: standard

nateprich:
## Summary

- forward Muse funnel invocations to one disposable, loopback session process
- load the Project lazily inside the command context and reuse its live view without a cache
- preserve per-command API telemetry, including the first load, and document the measured reduction
- cover session reuse, telemetry, runner startup/shutdown, and the existing wall-clock harness

Closes #291.

## Tests

- `python3 -m pytest -q`
- `bash -n scripts/muse-review`


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "858e310c03e4bcaf46160ef2c14eded6595d55df",
  "reviewed_at": "2026-09-10T10:25:28.087961+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-10T10:25:29.834327+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #530, opened 2026-09-10
usage_snapshot reads each agent's own pool, and records the unmetered marker for Muse


nateprich:
Parent: #514.

What: in `heartbeat.py`, `usage_snapshot(agent)` reads the agent's own pool through `usage.read_agent(agent, time.time())` instead of the `read_claude()`/`read_codex()` branch that put Codex's meter on every non-Claude record. Metered providers keep the reduced-window shape `{window: {used_percent, resets_at}}` the idle gate needs; an unmetered provider (Muse on meta) records the explicit marker `{"source": <provider>, "unmetered": true}` rather than an empty window map. Still best-effort and never fatal: unreadable usage returns `None`. Then post the note on #431 the plan asks for (comment only): the polling coefficient was measured on the Codex meter; give the corrected attribution and confirm the retire-zcode direction holds.

Accept (fixtures only, no network — stub `usage.read_zai`, `usage.read_codex`, `usage.read_claude` with distinct `resets_at` values; never the live spool): `usage_snapshot("zcode")` carries the z.ai `resets_at` and no Codex value anywhere in the record; `usage_snapshot("muse")` records the unmetered marker; `usage_snapshot("codex")` and `usage_snapshot("claude")` keep the window shape; a reader that raises yields `None`; an unregistered agent yields `None`. The #431 note is posted and linked from the PR.

Scope: `heartbeat.py` and tests; one comment on #431.

Risk: standard

Issue #514, opened 2026-09-10
heartbeat.usage_snapshot records Codex's meter on every non-Claude agent, so zcode and Muse records carry the wrong provider's usage


nateprich:
# Plan: per-agent usage snapshots in heartbeat records

## What this is

heartbeat.py usage_snapshot(agent) currently reads usage.read_claude() for
agent == claude and usage.read_codex() for everything else. Every zcode and
Muse heartbeat record therefore carries the Codex OpenAI rate-limit reading
labeled as its own usage. The issue counts 137 of 241 zcode records carrying
the Codex seven_day resets_at (1789445736) instead of the z.ai value
(1789291561), and two decisions already paid for it: the zcode retirement
measurement on #431 took a +14-point review from the Codex 09:41 counter
glitch, and the polling coefficient was drawn from lagged within-run deltas
on the wrong meter.

The fix routes the snapshot through the existing per-agent dispatch:
usage_snapshot(agent) calls usage.read_agent(agent, now), so zcode records
z.ai quota, Muse records its explicit unmetered shape, and claude and codex
keep their current readers. A regression test pins a zcode snapshot to the
z.ai resets_at and asserts it never carries the Codex value, and a short note
on #431 records that the coefficient was wrong but the direction held.

Proposed class: Broken

## Change

1. heartbeat.py usage_snapshot(agent): replace the
   read_claude()/read_codex() branch with usage.read_agent(agent, time.time()).
   Keep the function best-effort and never fatal: unreadable usage still
   returns None so a missing meter can never hide the run it documents.
2. Preserve the existing reduced-window shape for metered providers
   ({window: {used_percent, resets_at}}), which the idle gate needs to
   identify which five-hour window a run belonged to. For unmetered providers
   (Muse on meta), record the explicit marker {source, unmetered: True} rather
   than an empty window map, so the record says it was never gated instead of
   implying a budget was checked.
3. Add a regression test: stub read_zai and read_codex to return distinct
   resets_at values and assert a zcode snapshot carries the z.ai value and not
   the Codex value; assert a Muse snapshot carries the unmetered marker. Use
   fakes and a tmp spool only, never the live heartbeat spool.
4. Post a note on #431 (closed; comment only, no code change there) stating the
   polling coefficient was measured on the Codex meter, giving the corrected
   attribution from the 104 clean z.ai pairs, and confirming the retire-zcode
   direction still holds.

## Acceptance criteria

[...]

Issue #431, opened 2026-09-09
Retire zcode: Muse carries every standard-tier job, and a retired agent's silence must not alarm


nateprich:
Shaped 2026-09-09 by Claude on Nate's decision to retire zcode. One concern: **a retired agent must leave cleanly — no perpetual alarm, and a record that lets it come back.**

## The decision

Nate, 2026-09-09: *"kill those and just use muse going forward since muse is cheaper, and the models are better."* Measured before deciding (heartbeat, last 24h): zcode 93 runs — 18 did work (6 merges, ~10 breakdowns, 4 shapes), 63 refused on the z.ai pace line; since 09:50Z, one job in 34 runs, with its weekly pool sitting on the line (51.6% used vs 50.9% allowed). Muse in the same 24h: 217 runs, 98 did work (70 merges, ~26 breakdowns), zero budget refusals — and since #366 its standard schedule shapes standard-tier ideas, so it now carries every job `routines/zcode.md` describes.

## Decided from precedent

- **The Claude routine was retired the same way** — LEARNINGS.md, 2026-09-07: schedules deleted on Nate's instruction, cron expressions recorded in the repo because they live nowhere else, "what this leaves" stated.
- **The watchdog watches every agent in `heartbeat.PROVIDERS`** (`watchdog.py:151`) and infers each one's silence threshold from its own history (#35). zcode's history is a 15-minute cadence, so a paused schedule reads as "gone quiet" within about an hour — forever. That is the one code change retirement needs.
- **GitHub is the state; the heartbeat branch is the record.** zcode's records stay readable; nothing is deleted.

## Decided by the agent

- **One shared constant, `heartbeat.RETIRED_AGENTS`**, read by both the Actions watchdog and `agent_health` (the brief), so the two cannot disagree. `PROVIDERS` is untouched: `heartbeat.py`, `usage.py` and `funnel begin --agent zcode` keep working, so re-enabling is a schedule paste and removing one name from the set.
- **The routine file stays**, headed RETIRED with the date and reason, because the sha tests and the guardrail tests read it and because it is the re-enable instructions. `usage.py`'s `zai` policy stays, dormant; #95 is closed as moot.
- **Schedule recorded**: every 15 minutes at :08, :23, :38, :53, on the zcode app, first run 2026-09-07 02:11Z — from the heartbeat, since the app holds the schedule where nothing can read it (#52).

## Needs you

[...]

PR #541, opened 2026-09-10
Read heartbeat usage from each agent pool

nateprich:
## Summary

- make heartbeat usage snapshots call usage.read_agent(agent, time.time()) so zcode reads z.ai and each metered record keeps the idle-gate window shape
- record Muse as {source: meta, unmetered: true} and keep unreadable or unknown usage best-effort
- add fixture coverage for each provider, unmetered Muse, reader exceptions, and unknown agents
- post the corrected attribution note on #431: https://github.com/nateprich-projects/command-center/issues/431#issuecomment-5616860138

Closes #530.

## Verification

- python3 -m pytest tests/test_heartbeat.py — 28 passed
- python3 -m pytest — 933 passed; the two existing host-only Muse review-bound timeout tests remain tracked in #482
- git diff --check — clean

Deliberately unchanged: usage.py provider policy, idle-gate behavior, schedules, Status/Class fields, and the zcode retirement decision. No merge performed.

PR #558, opened 2026-09-10
Record each agent's own meter on its heartbeat, not Codex's

nateprich:
Ticket #530 (project #514, `Broken`).

## What

`usage_snapshot()` read `usage.read_codex()` for every agent but Claude, so 137 of 241 zcode records carried OpenAI's windows as their own, and two measurements on #431 were drawn from the wrong meter. The snapshot now goes through `usage.read_agent(agent, now)`, the per-provider dispatch the pace gate already uses: zcode records z.ai's windows, Muse records the explicit `{"source": "meta", "unmetered": true}` marker rather than an empty window map that implies a budget was checked, Claude and Codex keep the reduced-window shape the idle gate needs, and an unreadable meter is still `None` — never another provider's reading. One consequence stated plainly: Claude's reading now prefers a fresh statusline cache and falls back to the local estimate, the same preference the gate itself applies, where before it took the cache at any age.

The plan's note on #431 (the polling coefficient was measured on the Codex meter; the direction holds) is posted with the merge and linked from the ticket.

## Verification

- `tests/test_usage_snapshot.py`: a zcode record carries the z.ai `resets_at` and never Codex's; a Muse record says it was never gated; Codex and Claude keep the window shape; an unreadable meter is `None`, never another provider's; an unregistered agent is `None`. Fakes only; the live spool is untouched.
- Full suite locally: everything passes except `tests/test_muse_review_bound.py`, which times out on this loaded host (#482) and is unrelated.


nateprich:
Superseded: Codex took over #530 after this session's claim went stale without a branch for thirty minutes (#396's rule, working as designed), and its PR #541 merged with the same design — `usage.read_agent`, the unmetered marker, best-effort nulls. Nothing here is missing from main. Closed by Claude in session; the branch is left in place.

Issue #531, opened 2026-09-10
tests/conftest.py isolates the heartbeat for every test and fails on live-spool growth


nateprich:
Parent: #516.

What: add `tests/conftest.py` with an autouse fixture that points `heartbeat.SPOOL_DIR` at a per-test temporary directory and replaces `heartbeat.gh` with a stub raising `HeartbeatError("offline in tests")`, so every real `heartbeat.append()` spools to the temporary directory and nothing reaches GitHub; a test-level `monkeypatch` still overrides it (the watchdog and doctor spool tests keep their own isolation and pass unchanged). The same fixture records the live spool files' sizes under `~/.claude/command-center-heartbeat` before each test and fails the test, naming the file, if any grew. Then remove the three fake `zcode` rows (run `run-id`, `skipped-api-reserve`, 2026-09-09 04:34Z) from the `heartbeat` branch in one commit touching nothing else — a Mac-side step for the merger, recorded in the PR.

Accept (fixtures only, no network): a test that calls the real `heartbeat.append()` lands its record in the temporary spool and the live files are byte-identical after the suite; a deliberate write to the live spool inside a test fails that test with the file named; `tests/test_watchdog.py` and `tests/test_doctor.py` pass unchanged; the full suite passes on the Mac with the live spool untouched.

Scope: `tests/conftest.py`, tests; one cleanup commit on the `heartbeat` branch.

Risk: standard

Issue #516, opened 2026-09-10
The test suite writes real heartbeat records: begin and reserve-gate tests appended fake events and bind records to the live spool, and three reached the heartbeat branch


nateprich:
Shaped 2026-09-10 by Claude in session, from the spool pollution found the same morning.

## The defect

The test suite writes real heartbeat records. Running `tests/test_begin.py` on ticket/505 appended fourteen `bind` records with run `run-id` to the live spool at `~/.claude/command-center-heartbeat/{codex,muse,zcode}.jsonl`, because those tests stub `funnel.subprocess.run` for the push but not the file write inside `heartbeat.append()`. The remote `heartbeat` branch already carries three `skipped-api-reserve` events with run `run-id` in `zcode.jsonl` from 2026-09-09 04:34Z — the same hole through `heartbeat.record_event()` in the reserve-gate tests, and that time the push was real. The spool and the branch are the record of truth for `agent_health`, `unattended_merges`, the watchdog and the brief; fake rows pollute every one of them.

## The thing

- **A single test-wide isolation.** `tests/conftest.py` gains an autouse fixture that points `heartbeat.SPOOL_DIR` at a temporary directory and replaces `heartbeat.gh` with a stub that raises `HeartbeatError("offline in tests")`, so every real `append()` spools to the temporary directory and nothing reaches GitHub. Tests that exercise the spool or the push deliberately keep working: `test_watchdog.py`'s `_isolate_spool` and `_offline` already do exactly this, and a test-level `monkeypatch` overrides the fixture.
- **A guard that would have caught this.** The same fixture records the live spool files' sizes before each test and fails the test if any grew, naming the file — so the next writer that finds a way around the stub is caught at the test, not weeks later in the brief.
- **One cleanup.** The three fake `zcode` rows are removed from the `heartbeat` branch in a single commit that touches nothing else; the local spool was already cleaned by hand on 2026-09-10.

## Decided from precedent

- `test_watchdog.py` already isolates the spool per test; this lifts the pattern to the suite rather than inventing another.
- Instrumentation never gates the thing it instruments: the fixture stubs the transport, it does not make `append()` raise.
- Retired agents' rows are out of scope for the readers (#431), which is why the pollution stayed harmless in `agent_health`; it is not harmless in `unattended_merges` or the watchdog.

## Decided by the agent

[...]

PR #538, opened 2026-09-10
Isolate heartbeat writes in tests

nateprich:
## Summary

- add a function-scoped autouse fixture in `tests/conftest.py` that redirects `heartbeat.SPOOL_DIR` to a per-test temporary directory
- stub `heartbeat.gh` as `HeartbeatError("offline in tests")` so real appends spool without network access
- fail teardown when any live `~/.claude/command-center-heartbeat/*.jsonl` file grows, naming the file
- add a regression test for a real `heartbeat.append()` and byte-identical live files

## Verification

- focused: `python3 -m pytest tests/test_conftest.py tests/test_watchdog.py tests/test_doctor.py` — 111 passed
- full: `python3 -m pytest` — 930 passed; the two existing host-dependent `tests/test_muse_review_bound.py` cases timed out at the local 30-second subprocess bound (tracked in #482)
- `git diff --check` passes

## Heartbeat cleanup

The three specified fake zcode rows were removed from the `heartbeat` branch in standalone commit [`4774d41`](https://github.com/nateprich-projects/command-center/commit/4774d41), which is not part of this PR diff. No live spool files were written by the suite.

Deliberately unchanged: production heartbeat code, watchdog/doctor tests, Status/Class fields, and the live spool contents.

Issue #396, opened 2026-09-09
A claim with no ticket branch after 30 minutes is stale and taken over


nateprich:
Parent: #394.

What: `stale_locks()` gains a second rule beside the two-hour TTL — a claim older than **30 minutes** on a ticket with **no `ticket/<n>` branch** on the remote is stale, taken over like a TTL-expired one and counted in `stale_locks_taken_over`. Branch existence comes from the branch/PR facts the brief already fetches (`ticket_pr_facts`, bounded by #272); no new call per ticket. Thirty minutes is past the routine's clone-branch-first-push opening and far short of the hour that saturated the queue on 2026-09-09.

Accept (fixtures only, no network): 31 minutes and no branch → stale; 31 minutes with a branch → not stale; 20 minutes and no branch → not stale; the two-hour rule unchanged; a replay of 2026-09-09's five claims (#221, #214, #223, #224, #301) yields five takeovers. `in_motion` in the brief reflects the takeover.

Scope: `funnel.py` and tests. Do not change `WIP_LIMIT`, the TTL, or Broken preemption. A run that has pushed anything must never lose its lock.

Risk: escalated — concurrency; changes when a held lock is taken over

PR #560, opened 2026-09-10
Isolate the heartbeat spool for every test, and fail on live-spool growth

nateprich:
Ticket #531 (project #516, `Broken`).

## What

A test run on 2026-09-10 appended fourteen fake `bind` records to the live spool under `~/.claude/command-center-heartbeat`, and three fake events from an earlier run had reached the `heartbeat` branch, because tests stubbed the push but not the file write inside `heartbeat.append()`. `tests/conftest.py` now points `heartbeat.SPOOL_DIR` at a per-test temporary directory and replaces `heartbeat.gh` with a stub that raises `HeartbeatError("offline in tests")`, so every real `append()` spools to the temporary directory and nothing reaches GitHub. The same fixture records the live spool files' sizes before each test and fails the test at teardown — naming the file, a new file counting as growth — if any grew. A test-level `monkeypatch` still overrides it; the watchdog and doctor spool tests keep their own isolation and pass unchanged. `pytester` is registered from the conftest so plain `pytest -q` has it.

**Mac-side cleanup:** the three fake `zcode` rows were already gone from the `heartbeat` branch when the merger checked (511 lines, none with run `run-id`); the local spool was cleaned by hand this morning. Nothing else to remove.

## Verification

- `tests/test_conftest_isolation.py`: a real `append()` lands in the temporary spool; the transport is cut; a nested pytest run whose test writes into the (pytester-isolated) live spool path errors at teardown with the file named.
- Full suite on the Mac: everything passes except `tests/test_muse_review_bound.py` (#482, host-only timeouts); the live spool files are byte-identical before and after.


nateprich:
Superseded: Codex took over #531 after this session's claim went stale without a branch for thirty minutes (#396's rule, working as designed), and its PR #538 merged with the same design — a per-test spool directory, the transport cut, and the live-spool growth guard. Nothing here is missing from main. Closed by Claude in session; the branch is left in place.

Issue #557, opened 2026-09-10
Bound funnel session per-command work so one slow command cannot wedge the run


nateprich:
Parent: #553.

Depends on #556 — the per-command timings and named busy command from instrumentation are needed to bound the right path and prove the wedge is gone.

What: Bound server-side work per command so each command answers or fails loud inside the transport budget and frees the server for the next one. Document the in-run recovery when the session is wedged. No cross-run cache, no daemon, no re-begin re-claim path; begin claim-lock semantics unchanged.

Accept: One slow command cannot permanently wedge later commands in the same run — each command answers or fails loud inside the budget; begin claim-lock semantics unchanged (no cross-run caching, no second claim path); unit suite green.

Risk: standard

PR #580, opened 2026-09-10
Instrument funnel session health and command timings

### LINKED ISSUE #556: Funnel session: distinguish wedged from unreachable with per-command timings and health path
Parent: #553.

What: Split the funnel session client timeout error text into connect-timeout (session unreachable) versus reply-timeout (session busy past the 30s budget, naming the slow command when known). Add server-side per-command timings plus a cheap ping/health path that answers even when a worker is busy, so a run can tell wedged from dead.

Accept: After a successful begin, a slow or wedged session reports busy-with-named-command versus unreachable instead of a bare timeout; unit suite green.

Risk: standard

nateprich:
Fixes #556

## Summary
- run normal session requests in serialized worker threads so a health probe can answer while a command is busy
- track the active command and elapsed seconds, return per-command timing metadata, and retain the latest timing for health diagnostics
- probe the health path after a reply timeout so the client names the active slow command when known
- keep the existing 30-second transport budget and per-run FunnelSession semantics unchanged

## Verification
- `python3 -m pytest -q tests/test_funnel_session.py` — 16 passed
- `python3 -m pytest -q` — 969 passed
- `git diff --check` — passed

## Deliberately not changed
- no server-side command cancellation, timeout bounding, or in-run recovery; those belong to the dependent #557
- no cross-run cache, daemon, re-begin claim path, Status/Class changes, or live automation changes

The branch is pushed at commit `1378771`; this PR is intentionally not merged.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "137877120275fc1be5e37cdad0b7a02743727aab",
  "reviewed_at": "2026-09-10T16:58:12.808123+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-10T16:58:15.296763+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #615, opened 2026-09-11
Fail loud on degraded record-bearing brief sections


nateprich:
Parent: #613.

What: When closed_itself, cleared_blocks, or unattended_approvals exceeds its section budget, make funnel brief fail that section loudly (non-zero exit or explicit section-failed marker) and withhold the empty record list instead of printing `[]`.

Accept: A test forcing a section timeout asserts no empty record list is printed for the degraded section and the failure is explicit (non-zero exit or section-failed marker); existing brief tests still pass.

Risk: standard

Issue #165, opened 2026-09-08
Effective-dated rate table prices each outcome record in dollars


nateprich:
Parent: #15 — turning tokens into the dollar figure Nate asked for.

Added 2026-09-07. Nate: *"I want to start tracking API cost for each ticket — the amount this ticket would cost if we weren't using a personal subscription but rather connecting straight into the API."*

Two pieces:

1. **An effective-dated rate table.** One row per `(provider, model, token kind)` carrying the rate per million tokens, the instant it took effect, the source it was read from, and when it was recorded. Checked into this repository as a data file, following the precedent that constants live in the repo — `usage.py`'s capacity figures, `heartbeat.py`'s `PROVIDERS` and `MODEL_SOURCES`.
2. **A priced field on each outcome record.** For every run joined in #100, multiply its four token counts by the rates that were live *at that run's timestamp*, and sum.

**Load-bearing rules:**

- **Effective dating is the requirement, not a nicety.** Nate asked for new pricing to apply "starting from the day and time that the new pricing took effect". A table with one current rate per model silently reprices all of history the first time a vendor changes a number — and the change would be invisible, because the old figures would simply be different.
- **Git history is not the effective date.** When a row was committed and when the rate took effect are different facts and will not match. `effective_from` is an explicit column.
- **It is notional, and must be labelled so everywhere it appears.** Nate is on subscriptions; this is not a bill. #88 makes the same point and it has already misled this project once, when a zcode run's ~94% cache-read figure was quoted as though it were consumption.
- **Python measures, the model proposes** (`plan.md`, #15). No model asserts a rate or a total.
- **Missing tokens produce null, never zero.** A run with no token counts has unknown cost, and unknown must not average in as free.
- **No composite number.** #102's rule applies here: cost is a named signal beside rework and intervention, never rolled into an efficiency score.

Out of scope: fetching rates automatically or noticing that one changed. That is #162, which consumes this table rather than replacing it. This ticket seeds the table by hand from published rates and records where each came from.

Acceptance: a fixture run priced against a table whose rate changed mid-history is priced at the rate live at its own timestamp, not the newest one; a run with null tokens yields a null cost and the job completes; the four token kinds are priced at four distinct rates, verified by a fixture where cache reads and fresh input differ.

Dependencies: #164 for the token counts, and #100 for the join and the model. Plan sequencing, not advisory: nothing in #15 starts before #76 and #80 (from #59) land.

Risk: standard

nateprich:
[...]

PR #628, opened 2026-09-12
Attribute outcome runs to session token usage

### LINKED ISSUE #100: Attribute credits from heartbeat usage to outcome records
Parent: #15 — half one: the credits field of the outcome record.

Join per-run usage from the heartbeat spool (`heartbeat.py read`) to the outcome records built by #99, so each record carries the credits consumed by the run that did the work — attempts matched to runs, runs matched to their usage entries.

- Read-only over the spool. The record stays derived data; nothing is asked of any run.
- Usage is populated 88/101 today, so the join must tolerate a run with no usage record: credits are null, not zero, and no number is ever invented. Python computes every number.

Acceptance: for a closed ticket whose run recorded usage, the record's credits equal the spool's usage for that run; for one without, the field is null and the job still completes.

Dependencies: #99 for the record shape. Plan sequencing, not advisory: nothing in #15 starts before #76 and #80 (from #59) land.

Risk: standard

---

## Amended 2026-09-07 — carry the model alongside the credits, and stop at credits

Nate asked for the model and effort level used on each ticket. Both are already recorded:
`heartbeat.py:354` `detect_model()` writes `provider`, `harness`, `model`,
`reasoning_effort` and `model_source` onto every run's heartbeat, read from the agent's own
session file rather than asked of the agent.

They therefore cost this ticket nothing extra. **The join this ticket already builds — run
to ticket, run to usage — is the same join that carries them.** Copy them onto the outcome
record beside `credits`, with the same rule: absent is null, never guessed, and
`model_source` travels with the value so a later reader knows it was observed rather than
declared.

**And a boundary, so it is not invented here.** This ticket stops at credits and tokens. It
does **not** convert them to dollars: an effective-dated rate table is its own ticket,
because a price that changes has to be recorded with the instant it took effect, and runs
before that instant must keep the rate that was live when they ran. Emitting a dollar
figure from a single current rate here would silently reprice all of history the first time
a vendor changes a number.

Acceptance, in addition to the above: for a closed ticket whose run recorded a model, the
record carries that model and effort verbatim from the heartbeat; for a run where
`detect_model` found nothing, both are null and the job still completes. No dollar figure
appears anywhere in this ticket's output.

nateprich:
Closes #100

## What changed

- Join closed-ticket outcome records to durable heartbeat ticket bindings.
- Read each bound run's own Codex, Claude, Muse, or zcode session transcript.
- Normalize fresh input, cache-read input, cache-write input, and output tokens per run, retaining null for unreadable kinds.
- Carry detected provider, harness, model, reasoning effort, and source metadata onto the record.
- Deduplicate replayed Claude messages, ignore Codex cumulative event messages, and extract Muse tokenUsage events from its binary journals.

## Deliberate boundary

This PR does not read heartbeat quota percentages as credits and does not invent a dollar value. The per-run timestamped token observations are the input for #165's effective-dated rate table and pricing.

## Verification

- Focused: 21 passed.
- Full: 1033 passed, 1 skipped.
- Read-only live derivation verified a real closed Codex ticket run (#615) with model, effort, and four token kinds.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "dd37284f6892f06adb4ae4ec38cb83cee6fac10f",
  "reviewed_at": "2026-09-12T21:03:07.226310+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-12T21:03:09.587702+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #694, opened 2026-09-13
Document console-only LaunchAgent bootstrap handoff

### LINKED ISSUE #690: Restore keeper LaunchAgent bootstrap from the automation shell
Parent: #689.

What: On the Mac that runs the schedule, diagnose why launchctl bootstrap into the gui domain fails with error 5 from the automation shell, then restore the documented reinstall path. Record the execution context first (how the shell was entered, Aqua login session state, launchctl print output for the keeper label). Reproduce from both contexts (automation shell and an interactive Terminal on the console), comparing bootstrap of the installed keeper copy against a minimal RunAtLoad probe. Check the mechanical causes: installed copy byte-identical to the repo copy, ownership and permission bits on the installed file and its log paths, and the exact bootstrap domain used (plist header comment gives the exact bootout plus bootstrap commands; scripts/install.sh is the documented alternative). Fix at the diagnosed cause: if the gui domain is unreachable from the automation shell, document and script the supported reinstall context (console login session) and align the drift-test failure message or install script notes so the next run is not handed a command that cannot work where it runs; if the evidence names file content instead, fix the canonical repo copy and re-copy repo to installed. Do not symlink the installed plist (measured 2026-09-07: launchd refuses it with Bootstrap failed 5), do not sync installed to repo, do not edit the installed copy without bootout plus bootstrap, and do not weaken the drift test. Coordinate with the sibling keeper-drift ticket work in flight (#670, #635): the drift test already reads green here, so do not re-sync the same plist file twice unless the evidence names content as the cause.

Accept: the keeper job shows loaded in launchctl print, the whole tests/test_launchd_drift.py file is green, and a fresh keeper heartbeat record is appended.

Risk: standard

nateprich:
## Summary

Closes #690.

- Record the measured distinction between copying LaunchAgent plists and bootstrapping them: the managed automation shell can inspect `gui/501` and boot out a job, but `launchctl bootstrap` returns `Bootstrap failed: 5`.
- Make `scripts/install.sh` print an explicit console/Aqua Terminal reload handoff instead of implying that copying reloads launchd.
- Make the launchd drift assertion message point to the same console-only handoff and add regression coverage for the guidance.

## Evidence

- `launchctl print gui/501` showed `session = Aqua`; the keeper was initially loaded and the installed plist was byte-identical to the repo copy, plist-valid, user-owned, mode `0644`, with valid user-owned log paths.
- After the documented bootout, bootstrap from the non-TTY automation shell returned `Bootstrap failed: 5: Input/output error`. A minimal `RunAtLoad` plist and a PTY-backed terminal shell returned the same error.
- `launchctl kickstart gui/501/com.nateprich.command-center-run-keeper` restored the already-registered unchanged keeper; `launchctl print gui/501/com.nateprich.command-center-run-keeper` then succeeded. The fresh sentinel record was timestamped `Sat_Sep_12_20:50:03_2026`.
- The native Terminal app was not accessible through this agent UI surface, so the PTY result is reported separately rather than claimed as a console-Terminal run.

## Deliberately not done

- Did not change the canonical plist, installed plist, schedule, or drift-test strictness.
- Did not replace `bootstrap` with `kickstart`: kickstart only recovered the unchanged already-registered service and is not a safe reload for a changed plist.

## Verification

- `PYTHONDONTWRITEBYTECODE=1 python3 -m pytest -q tests/test_launchd_drift.py` — 25 passed, 1 skipped
- `bash -n scripts/install.sh`
- `git diff --check`

The installed keeper is loaded and the drift suite is green on this host. Please do not merge automatically; the supported reload handoff now requires a logged-in console/Aqua Terminal session.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "ac744805bec5a808c49fbfd94f12a011f6120670",
  "reviewed_at": "2026-09-13T03:57:15.704841+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-13T03:57:15.705076+00:00",
  "run": "d78ffb94557a",
  "voice": "agent"
}
```

Issue #775, opened 2026-09-13
Load the funnel publisher LaunchAgent on the schedule Mac


nateprich:
Part of #643; discovered while implementing #652.

The publisher job ships as code in the #652 PR, but launchd only runs it after a console load. From a Terminal in the logged-in console (Aqua) session, not an automation shell:

  cp launchd/com.nateprich.command-center-funnel-publisher.plist ~/Library/LaunchAgents/
  launchctl bootout gui/$(id -u)/com.nateprich.command-center-funnel-publisher 2>/dev/null || true
  launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.nateprich.command-center-funnel-publisher.plist
  launchctl print gui/$(id -u)/com.nateprich.command-center-funnel-publisher

Then confirm one tick ran:

  tail -5 ~/Library/Logs/command-center-funnel-publisher.err.log

Also confirm the run-clone .env holds CLOUDFLARE_API_TOKEN and CLOUDFLARE_ACCOUNT_ID (#648 placed the token; publisher.py reads the run-clone .env by default). Until #650 lands there is no spool to publish, and until #653 lands there is no namespace; both show up as log lines, not as failures elsewhere.

Human step: a Claude Code environment

Risk: standard

nateprich:
**Installed and loaded** by the funnel watch (Claude Code desktop scheduled task on the schedule Mac), 2026-09-13 16:29 PDT, heartbeat run `c5abc208666d`.

1. **Precondition:** the runner is on `main`: `publisher.py` and the plist landed in #776 (f1f641e), with the env path fix in #786 (e28bd5e). The run clone is clean at `c5c6655` = `origin/main`.
2. `cp launchd/com.nateprich.command-center-funnel-publisher.plist ~/Library/LaunchAgents/`, then `launchctl bootstrap gui/501 …funnel-publisher.plist`: **rc=0**. It bootstrapped from this session; a console Terminal was not needed. It was not loaded before, so there was nothing to boot out.
3. `launchctl print`: `runs = 1`. `launchctl list`: `-	1	com.nateprich.command-center-funnel-publisher`. `RunAtLoad` fired one tick, and `StartInterval` 60 continues it.
4. **The first tick** (`command-center-funnel-publisher.err.log`):
   - `publisher: no spool entries yet; nothing to publish`: expected until #650 lands.
[...]

Issue #756, opened 2026-09-13
Instrument funnel brief by stage and report where the 179s goes


nateprich:
Parent: #750.

What: add per-stage timing to the funnel brief path (Project load, GraphQL calls by operation, brief assembly) and run it at least twice against current Project state; report per-stage timings and name the dominant cost.

Accept: a comment or PR description with per-stage timings from at least two runs on current state plus the identified dominant cost; output semantics of brief unchanged; existing brief tests still pass.

Risk: standard

nateprich:
Diagnosis complete; implementation is in PR #763. Two fresh brief runs on 2026-09-13 both returned JSON within the 179s command budget (45.89s and 45.19s wall time). Project load was 22.552733s and 22.590353s; its 11 `project_items` GraphQL calls took 20.727564s and 20.931872s. Membership took 1.020611s and 0.880689s. Three `brief_comments` calls took 3.939637s and 3.351246s. Brief assembly took 0.000015s in both. The dominant cost is the 11-page Project-item read, about 20.8s of the 22.6s Project load; #757 owns reducing it. Existing informational section timeouts remained visible and no gate semantics changed.



```json
{
  "agent": "codex",
  "at": "2026-09-13T13:19:38.978448+00:00",
  "run": "ddd6f871d004",
  "voice": "agent"
}
```

nateprich:
Second board-state sample for #750 instrument step (UTC; run 16cda8093e28). The captured #762 spend pair is 13 GraphQL calls / 35 points and 15 GraphQL calls / 37 points, with more than 3300 points remaining in both attempts. That makes GraphQL point exhaustion an unlikely dominant cause.

The post-cutoff transient check is clean at full-brief level: fresh checks started 2026-09-13T14:43:33.554481Z and 2026-09-13T14:44:33.024981Z, returned exit 0 with JSON in 47.09s and 45.88s, and did not reproduce a 179s command-timeout. Each current check used 16 GraphQL calls / 38 points, leaving 4415 and 4319 points respectively.

Both current checks still hit the bounded 8s ticket_pr_facts read and marked stranded/in_motion/stale_locks_taken_over missing; the gate-feeding sections completed. Verdict: the original full-brief 179s timeouts were transient, while the recurring bounded ticket_pr_facts degradation remains for the existing #757 fix / #758 guard line to consider. No brief code changed in this check.



```json
{
  "agent": "codex",
  "at": "2026-09-13T14:47:32.131119+00:00",
[...]

Issue #653, opened 2026-09-12
Auto-deploy dashboard on main changes


nateprich:
Parent: #643.

Depends on #648 - the first deploy needs the scoped token, per the plan.

What: Add a launchd job in the run-keeper mould that runs npx wrangler deploy when main changes under dashboard/, using the same scoped token from .env. Include KV namespace creation and the first deploy in this ticket, plus the funnel.nateprich.com route wiring.

Accept: A merged change under dashboard/ is deployed without a keyboard; the first deploy serves the Worker page; redeploys only follow dashboard/ changes.

Risk: escalated - deploys with the Cloudflare credential

Issue #652, opened 2026-09-12
Launchd publisher pushes spool snapshots on refresh demand


nateprich:
Parent: #643.

Depends on #648 - the Cloudflare token must exist in .env before the publisher can push anything.

What: Add a launchd publisher job on the Mac holding the only Cloudflare credential. Every minute it pushes the newest spool entry if unpublished. It also reads the refresh flag from Cloudflare storage; when the flag is set and the newest snapshot is more than 10 minutes old, it runs funnel.py brief from the run clone once, then clears the flag. Read the token from the gitignored .env; send failures to its own log and never touch agent runs.

Accept: With fixture spool entries and a test flag, the job publishes the newest entry once, runs one brief per set flag only when the snapshot is stale, clears the flag, and logs failures without affecting any brief exit code.

Risk: escalated - handles the Cloudflare credential

PR #776, opened 2026-09-13
Launchd publisher pushes spool snapshots on refresh demand (#652)

nateprich:
Implements #652, step 3 of the #643 dashboard plan.

## What

- **publisher.py** (new, stdlib-only, 3.9-compatible): one tick pushes the newest spool entry to the `snapshot` KV key when the remote copy differs, then polls `refresh-requested`. A set flag runs `funnel.py brief` from the run clone once when the newest snapshot is older than 10 minutes, then clears the flag. Everything logs to stderr (the job\u2019s own log under launchd); exit 0 on a completed tick even when the brief subprocess fails, so the publisher never affects any brief exit code.
- **tests/test_publisher.py** (new, 24 tests): fake in-memory KV server, fixture spool entries, and a fake brief script. Covers publish-once, newest-selection, unparseable-entry tolerance, refresh stale/fresh/unset, failed-brief still clears, KV-error exit 1 with no brief run, token-reaches-KV-but-never-the-log, overlapping-tick skip, brief timeout, and a subprocess probe that the module imports neither funnel nor heartbeat.
- **launchd/com.nateprich.command-center-funnel-publisher.plist** (new): every-minute poll, own log files, pinned `/usr/bin/python3` running the run-clone copy.
- **scripts/install.sh + tests/test_launchd_drift.py**: installer copies the plist; drift, wellformedness, exact-args, every-minute, and own-log tests cover it.

Branch: new `ticket/652`; no prior remote branch or PR existed (the `prior_run.py` match was heuristic noise for #756), so nothing was continued or reset.

## Departures and judgement calls (review against #643)

1. **#650 (spool writer) is still open, so this PR defines the spool read contract first**: dir default `~/.claude/command-center-dashboard-spool` (override `COMMAND_CENTER_DASHBOARD_SPOOL`), one JSON object per brief with at least `generated_at`, pushed byte-for-byte so extra keys ride along. A missing dir reads as an empty spool. Coordinated on #650; if its writer lands elsewhere the publisher default is a one-line change.
2. **A set flag on a fresh snapshot is cleared without running.** The ticket specifies only set+stale; the plan says a fresh tap "does nothing", and leaving the flag set would fire a delayed brief for a forgotten tap. Clearing without running honours both.
3. **Publish dedup compares against the remote snapshot** (byte-equality, then `generated_at`) instead of a local published-state file. No extra state to lose, and a wiped KV key self-heals on the next tick. A remote snapshot newer than local is left alone with a warning, never regressed.
4. **StartInterval 60, not the keeper\u2019s 60-entry calendar schedule.** A poll loop does not need drift-free firing, and overlap is handled by an in-script lockfile: the brief budget (120s) exceeds the tick (60s), so without the lock one flag could run two briefs. A tick that cannot lock fails closed.
5. **Namespace id is read from `dashboard/wrangler.toml`** (the file #653 already owns) rather than a second config; account id and token use wrangler\u2019s `CLOUDFLARE_*` names in `.env` so #653 shares the file with zero new config.

## Deliberately not done

- Spool writer (#650, open) and KV namespace + first deploy (#653, open). Pre-#650 ticks log "no spool entries yet"; pre-#653 KV calls fail into the publisher log only.
- Loading (bootstrap) the LaunchAgent: console step, filed as #775 under #643.

## Unsure about

- **`.env` location.** The publisher defaults to the run-clone root `.env` (sibling of the script). #648 put the token "in the gitignored .env on the Mac"; if Nate placed it in the working tree instead, #775 aligns the file or the override at load time.

## Verification

- `tests/test_publisher.py`: 24 passed; with drift tests 60 passed, 1 honest skip (plist not installed here).
- Full suite: 1189 passed, 1 skipped, 1 failed in `test_automation_drift.py`, confirmed failing on untouched `main` (pre-existing Codex automation drift, tracked by #163).
- Canonical in-memory compile of `funnel.py` and `publisher.py`; dashboard JS checks 10/10 (untouched by this change; ran `node --check` + `node --test` directly because the `npm` wrapper stalls in the sandbox).


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "6748eae9168e729ef3d69081772920aa9b59a5f3",
  "reviewed_at": "2026-09-13T17:11:46.365031+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-13T17:11:48.191016+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #789, opened 2026-09-13
Installed Codex automation prompts drift from routine


nateprich:
# Installed Codex automation prompts drift from routine — verify and close against #778

## What

#789 is a duplicate observation of drift already owned inside the funnel: on ticket/164 the canonical full suite failed tests/test_automation_drift.py with the same four installed Codex automation prompts stale against routines/codex-work.md (command-center-tickets-mon-fri-after-midnight, command-center-tickets-sun-thu-late-night, command-center-tickets-weekday-mornings, command-center-tickets-weekend-early-mornings).

The owning work is #778, sub-issue of the #770 resync project. Its 2026-09-13 comment records the identical state from the schedule host: one prompt fixed (command-center-tickets-hourly), the same four still drifted after the sync failed with [Errno 1] Operation not permitted creating automation.toml.bak, and the drift test at 1 failed, 20 passed. The prescribed remaining step is already recorded there: rerun the sync from a session with write access to those four installed automation directories, then close #778 after the drift test is green.

So this plan builds nothing. It scopes #789 to verify-and-close:

1. On the schedule Mac, from the canonical checkout, run `python3 scripts/sync_codex_automations.py --check` without changing anything and confirm the drifted names are exactly the four recorded on #778.
2. After #778 lands its rerun, rerun `PYTHONDONTWRITEBYTECODE=1 python3 -m pytest -q tests/test_automation_drift.py` to green.
3. Close #789 with a pointer to #778; #778 owns the fix and its own close-out.

Out of scope: any rewrite of installed prompts under #789, any change to routines/codex-work.md, schedules, models, status, tier and presence derivation, and any new sync machinery. Touch no repository files under this plan.

## Decided from precedent

- The routine file is the source of truth and the installed copies follow it. Source: scripts/sync_codex_automations.py docstring and tests/test_automation_drift.py docstring, as recorded in #770.
- Only the prompt field of the installed automation.toml copies is ever rewritten; schedule, model, status and everything else stay as the app wrote them. Source: sync script docstring and the #778 ticket body.
- Done means the drift test passes with zero drifted of the installed set. Source: the test assertion.
- The check is meaningful only where the automations directory exists; elsewhere the test skips, and absence is not evidence of no drift. Source: the test skip behavior quoted in its docstring.

## Decided by the agent

[...]

Issue #778, opened 2026-09-13
Resync the five installed Codex automation prompts from routines/codex-work.md


nateprich:
Parent: #770.

What: On the Mac where the Codex schedules live, from the canonical checkout, confirm the drift without changing anything (`python3 scripts/sync_codex_automations.py --check`), rewrite the installed prompt fields (`python3 scripts/sync_codex_automations.py`), then rerun `tests/test_automation_drift.py` to green plus the sync related tests around it. Touch only the `prompt` field of the installed `~/.codex/automations/*/automation.toml` copies; leave schedule, model, status, `routines/codex-work.md`, and all sync machinery unchanged.

Accept: `tests/test_automation_drift.py` passes with zero drifted of the installed set on the schedule Mac, and the sync related tests around it pass.

Risk: standard

nateprich:
Observed on the schedule host during this run:

- Initial `python3 scripts/sync_codex_automations.py --check` reported all five installed Codex prompts drifted from `routines/codex-work.md`.
- `python3 scripts/sync_codex_automations.py` updated only `command-center-tickets-hourly`. The other four attempts failed before their prompt writes with `[Errno 1] Operation not permitted` while creating `automation.toml.bak` under their installed automation directories.
- A fresh `--check` reports 4/5 still drifted: `command-center-tickets-mon-fri-after-midnight`, `command-center-tickets-sun-thu-late-night`, `command-center-tickets-weekday-mornings`, and `command-center-tickets-weekend-early-mornings`.
- `PYTHONDONTWRITEBYTECODE=1 python3 -m pytest -q tests/test_automation_drift.py` reports `1 failed, 20 passed`; the failure is the expected four remaining installed prompts.

No repository files, schedules, models, status, class, or sync machinery were changed. One installed prompt field was corrected; four remain blocked by this Codex sandbox. Please rerun the sync from a session with write access to those four installed automation directories, then close #778 after the drift test is green.



```json
{
  "agent": null,
  "at": "2026-09-13T18:45:18.662215+00:00",
  "run": null,
  "voice": "agent"
}
```

nateprich:
Done from the funnel watch (Claude Code desktop scheduled task on the schedule Mac, 2026-09-13 16:2x PDT, heartbeat run `c5abc208666d`). A Codex run cannot write `~/.codex/automations` from its sandbox (the `[Errno 1] Operation not permitted` above), so the watch did the step, per its playbook.

[...]

PR #790, opened 2026-09-13
Record per-run token usage in heartbeat finishes

### LINKED ISSUE #164: Heartbeat records per-run token totals, split by kind
Parent: #15 — the missing input for a dollar figure.

Added 2026-09-07 after Nate asked for API cost per ticket. Building it exposed a gap: **the heartbeat does not record token counts.** `usage_snapshot()` (`heartbeat.py:431`) keeps `used_percent` and `resets_at` per window — percentages of a subscription quota, not tokens. #100 joins that to outcome records and calls it credits, which is correct for budgeting and useless for pricing.

Make `heartbeat finish` record the run's own token totals, broken out by kind:

- fresh input
- cache read
- cache write
- output

**The four kinds are not optional detail; they are the whole point.** #88 measured cached input as the largest single line item for every model — Luna re-sent its context 16x, Sol 22x, a zcode breakdown run 27x — and cache reads are priced at a fraction of fresh input. A single blended token count prices the run wrong and hides exactly the effect #88 exists to watch.

Read them the way `detect_model()` already reads the model: from the agent's own session file, the same paths `MODEL_SOURCES` and `usage.py` already glob. Sum the run's own records, not the file's whole history.

**Load-bearing rules:**

- **Detected, never self-reported.** No new optional flag on `heartbeat finish` for an agent to pass. The schema-rich/data-empty table in #15 is what that produces — `ci_green` at 2/101.
- **Instrumentation must not gate the thing it instruments** (`plan.md`). Unreadable tokens are recorded as null and the run continues; this must never raise.
- **Capture at finish, not at report time.** Rejected alternative: leave the transcripts alone and have #99's walker read them when it derives the record. Transcripts are local-only and explicitly "buffers and recovery aids, never state of record" (`plan.md`, Architecture) — a cost figure that silently degrades as Claude Code prunes old sessions is worse than none. Capturing at finish puts the numbers on the heartbeat branch, where they are in GitHub and stay re-derivable.
- **The consequence, stated plainly:** history before this lands can only be priced as far back as local transcripts survive. #101's backfill must report how many records it could not price rather than emitting zeros.

Acceptance: a finished run's heartbeat record carries all four token counts for its own session; a run whose session file cannot be read records four nulls and still finishes; fixture tests cover at least Claude and Codex transcript shapes, and the muse snapshot shape if its token counts are reachable.

Dependencies: none — this is upstream of #100 and can start alongside #99. Plan sequencing, not advisory: nothing in #15 starts before #76 and #80 (from #59) land.

Risk: standard

nateprich:
Closes #164.

## What changed
- Add a durable `token_usage` snapshot to every normal heartbeat finish, with fresh input, cache-read input, cache-write input, and output token counts.
- Read the heartbeat-bound session through the existing Codex/Claude/Muse parser, bounded by the run timestamps; unreadable transcripts and missing token kinds stay null and never block a finish.
- Prefer the durable finish snapshot when deriving outcomes; keep transcript parsing only for pre-#164 heartbeat records.
- Add finish/outcome regression coverage and document the durable source.

## Deliberately not changed
- No dollar pricing, rate table, quota gate, or provider-meter behavior.
- No installed automation, launchd state, schedule, or credential changes.
- No changes to the existing legacy `input_usage` resend diagnostic.

## Verification
- Canonical compile passed.
- Focused telemetry/heartbeat/outcome suite: 80 passed.
- Full suite: 1,200 passed, 1 skipped.
- Full suite still reports the pre-existing installed Codex automation drift failure: 4 of 5 installed prompts do not match `routines/codex-work.md`; this is tracked by #778 and was captured from this run as #789.

The PR is intentionally left open for review.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "4ee3a147fc0705455bceb94b479c2311983e395a",
  "reviewed_at": "2026-09-13T21:58:23.310624+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-13T21:58:25.406698+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #839, opened 2026-09-14
Resolve implement-workspace uid so the SSH push path works end to end


nateprich:
Parent: #838.

What: In a fresh Muse implement workspace, first reproduce the reported failure (record `id`, the system user-database lookup for the active uid, `ssh -T git@github.com`, and a push of a scratch ref over SSH), then fix the workspace setup at one choke point so the active uid resolves to a user before any git or ssh step runs and the SSH transport to the forge works as provided. Fix at the workspace-setup layer, not with per-call fallback or retry handling around push. Leave no checkout-local `insteadOf` rule behind in the routine path; the supported path under test stays SSH.

Accept: the uid resolves via the system user database; a non-interactive `ssh -T git@github.com` reaches the forge; a push of a scratch ref over SSH succeeds and the scratch ref is then removed; no checkout-local `insteadOf` rule remains in the routine path.

Risk: standard

PR #853, opened 2026-09-14
Add SSH workspace regression coverage

### LINKED ISSUE #840: Regression coverage for workspace uid resolution and the SSH push path
Parent: #838.

Depends on #839 — the regression test must pass against the fixed workspace setup, so the fix has to land first.

What: Add regression coverage that fails before the #839 fix and passes after, covering uid resolution and the SSH push path in a test workspace. Verify the fail-before by running the new test against the pre-fix setup (e.g. a workspace without the fix applied), not only against the fixed one.

Accept: the new test fails on the pre-fix setup and passes on the fixed setup.

Risk: standard

nateprich:
## Summary

- Add disposable-workspace integration coverage for the Muse implementer.
- Resolve the active uid through the system user database.
- Exercise the SSH handshake and a real Git push/delete against a local bare remote.
- Assert the workspace uses an SSH origin and has no checkout-local `insteadOf` rewrite.
- The fake Muse boundary refuses the probe without `--disable-sandbox`, so the test fails against the pre-fix runner.

## Verification

- Pre-fix check: temporarily removed `--disable-sandbox`; the new test failed with the modeled nested-sandbox error (exit 42); restored the flag.
- In-memory `funnel.py` compile passed.
- Focused Muse suite: 8 passed.
- Full suite: 1,283 passed, 1 skipped.
- `bash -n scripts/muse-implement` and `git diff --check` passed.

## Scope

This PR adds tests only; the production `--disable-sandbox` fix is already on `main` from #839.

Closes #840.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "690dca80ea55505c53fc19dbe3ecd3849a5e7f5a",
  "reviewed_at": "2026-09-14T07:57:38.461796+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-14T07:57:40.145203+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #810, opened 2026-09-14
shape-packet and shape-apply with the plan schema and mechanical self-approval


nateprich:
Parent: #794.

Depends on #798 — must land first.

What: shape-packet <idea> prints the idea, origin, plan.md, AGENTS.md and sibling plans. shape-apply reads {decided_from_precedent[{claim, source}], decided_by_agent[{decision, alternative, why}], needs_nate{exposure, gates, scope, preference}, proposed_class, plan_markdown}. Render the body from fields; apply the self-approval rule in code (all four null, class self-approvable, origin agent); write Shaped or Ready; write the self-approval record as a code-owned comment. Fixture tests.

Accept: tests/test_engine_shape.py passes; a plan with one non-null needs_nate stays at Shaped in fixtures; the old Needs-section parser is not called.

Risk: escalated

Issue #794, opened 2026-09-14
Runner executes the protocol, the model answers questions: a deterministic engine for every scheduled job


nateprich:
# Plan: runner executes the protocol, the model answers questions

## The problem

The five routine prompts and three skills total about 31,000 words. Roughly 85 percent is
protocol: which command to run next, what to do with each JSON result, which outcome code
to record, how to write a marker a regex will later find. A model executing a protocol from
prose is the non-determinism. `funnel.py` then carries 18 compiled regexes and about 20
marker constants whose only job is to parse the text the model wrote, and when either side
moves the other breaks. At least 23 of the 89 Broken projects filed since launch attribute
themselves to an earlier change. Fixes land as prose, prose must be re-pasted into three
vendors' apps, and a run on the stale copy re-triggers the defect.

Broken stays first on the ladder. Its rank is not the cause. Where fixes land and who
executes them is.

## The change, in one sentence

**The model never runs `funnel.py` or `gh`.** For each job it receives a packet, returns
one structured answer, and a runner validates that answer against a schema and performs
every side effect.

Stated precisely: for **review** this is enforced, because Muse runs with
`--disable-shell --disable-write --disable-web-tools` and the packet is inline, so the
model has no tool at all. For **implement** the model needs a shell to run tests, so the
runner makes `funnel.py` and `gh` unnecessary rather than impossible. For **Codex** the
model has a full shell by construction. The guarantee is strongest exactly where a
mistake merges.

## What stays exactly as it is

- `funnel.py`'s ordering, ladder, gates, lock, budget gate, merge conditions, and the
  fixture tests on them. These are already deterministic and are the part that works.
- GitHub as the only state. The Project's `Status` and `Class`. The two-label vocabulary.
- The scheduling boundary: Muse under launchd, Codex in its app, Claude never headless.
- Nate's gates: accept on `New` and `Replace`; the Shaped gate when a plan holds an open
  question; parking. The 2026-09-10 decision that finite-class agent plans self-approve.
- `plan.md` as the design record, `AGENTS.md` as the rule book. Both get shorter, not
  different.

## The five jobs and their contracts

Each job has a **packet** the runner assembles, a **question** the model answers, an
**answer schema** the runner validates, and **effects** the runner performs. The model's
prompt is the question plus the judgement text for that job and nothing else.

### Review

| | |
|---|---|
[...]

PR #854, opened 2026-09-14
shape-packet and shape-apply with the plan schema and mechanical self-approval (#810)

nateprich:
Part of #794 (Phase 2). Implements the Shape row: packet carries the idea, origin, plan.md, AGENTS.md and sibling plans; apply validates the answer, renders the body from fields, and self-approves mechanically.

Fresh branch: no remote ticket/810 or PR existed, so nothing was continued or reset.

What it does:
- engine/shape.py: shape-packet (read-only collect + pure build_packet) and shape-apply (validate_answer, render_plan, decide, apply_shape). Entry points shape-packet / shape-apply mirror review-packet.
- The Shaped/Ready decision feeds needs_nate fields into the shared self_approval_eligible predicate. The Needs-section parser (shaped_plan_status / plan_needs_nate) is not called anywhere on this path; tests patch it to raise and prove the fields rule over prose in both directions.
- Ready posts the Self-approved record as a code-owned comment; one non-null needs_nate stays at Shaped with no comment.
- 57 fixture tests in tests/test_engine_shape.py, mirroring tests/test_engine_review_packet.py.

Deliberate choices (flagged plainly):
- Kept the escalation term: decide() also requires no escalated risk via plan_is_escalated on the rendered plan. The ticket parenthetical lists three conditions; the fourth keeps the existing gate (an escalated-risk plan must not self-approve to Ready). #794 says the self-approval rule is applied mechanically, and escalation is part of that rule.
- Override blocks are carried over verbatim and read from the original body; cmd_shaped drops them, which would make the override term dead on this path while the packet reports it. This is the one behavioral departure from funnel shaped.
- Kept cmd_shaped parity otherwise: authority-signal advisory in the basis, overlap advisory print, class recovery write immediately before the status write, unconfirmed-status path.
- Proposed class renders ahead of the decision record so ## Needs Nate stays last and parser-clean for readers that still parse it through the cutover.
- No funnel.py, routine, skill, or plan.md changes: purely additive (4 new files).

Deliberately not done: shadow runner, plist, cutover, deleting the old parser (later Phase 2 tickets); no Needs-heading check on plan_markdown (fields are authoritative per the plan\u2019s structured-output tolerance).

Unsure about: whether the Phase 2 cutover wants the escalation term to stay a prose scan or become an answer field; left as a scan for gate parity.

Verification: 57/57 new tests pass; full suite 1339 passed, 1 skipped, plus 2 teardown errors from the live heartbeat spool growing mid-run (pass in isolation, unrelated: this diff touches no existing file); shape-packet run live on #810 prints valid JSON with all fields (12 sibling plans).


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "e749828011eebcf740e0a311cd7daf98b4c61cd6",
  "reviewed_at": "2026-09-14T13:45:43.668224+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-14T13:45:45.859430+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #815, opened 2026-09-14
implement-packet and finish-ticket, happy path


nateprich:
Parent: #794.

Depends on #798 — must land first.

What: implement-packet <ticket> prints ticket, plan, the newest verdict's blocking list, and the prior-run digest. finish-ticket runs from a clone on ticket/<n>: run the tests, commit, push, open the PR from a template that carries the model's summary and departures, release the claim, finish the heartbeat with an outcome derived from what happened. Fixture tests with a temporary git remote.

Accept: tests/test_engine_implement.py passes; finish-ticket on a fixture clone opens a PR and releases the claim.

Risk: escalated

Issue #806, opened 2026-09-14
Review cutover: swap the live review plists to the engine after the shadow period


nateprich:
Parent: #794.

Depends on #804, #805 — must land first.

What: After at least 48 hours or 50 shadow jobs, run shadow-report and record it on this ticket, including its list of disagreements (#977). Label every disagreement in the comment as one of: engine wrong (say whether it wrongly approved or wrongly rejected), old reviewer wrong, measurement error, or judgement call, with one line of evidence each. Only if engine-wrong disagreements are at most 5 percent of compared jobs, no engine-wrong disagreement is a wrong approval, and malformed output is below 3 percent: bootout the two old review plists, bootstrap the engine plist without --shadow for both tiers, bootout the shadow plist. If any threshold fails, or a disagreement cannot be labelled with evidence, stop and comment; do not cut over. Nate replaced the original 95 percent raw-agreement gate with this one on 2026-09-16: raw agreement also counted old-reviewer errors and measurement errors against the engine.

Accept: the shadow-report output and the labelled disagreement list on the ticket meet all three thresholds; launchctl list shows the engine review jobs and neither old review job.

Human step: a Claude Code environment

Risk: escalated

nateprich:
**Blocked on #901: the shadow report cannot answer this ticket's threshold, in either direction.** Run on the schedule Mac, 2026-09-15 16:3x PDT (funnel watch, heartbeat run `94a1a4f9172f`):

```json
"agreement": { "agree": 0, "disagree": 0, "rate": null,
               "approved": {"live": 0, "shadow": 2},
               "rejected": {"live": 0, "shadow": 2} },
"jobs": { "compared": 0, "live": 4, "matched": 4, "shadow": 4 },
"malformed_output": { "live": {"count": 0, "rate": 0.0},
                      "shadow": {"count": 0, "rate": 0.0} },
"time_per_job": { "live": {"median_seconds": 146.5, "p90_seconds": 205.2},
                  "shadow": {"median_seconds": 91.5, "p90_seconds": 134.2} }
```

Four shadow jobs, four live jobs, four matched by PR, **zero compared**: no live job carries a verdict the report can read, because the live reviewer is still the prose routine and its finish note says `merged PR #896`, while `--review-result` is passed by nothing in the tree. So `rate` is `null`, not low — the comparison is empty, and "agreement at or above 95 percent" cannot be established. Cause and fix are on **#901**.

[...]

Issue #805, opened 2026-09-14
shadow-report: agreement, malformed rate and time per job from the heartbeat


nateprich:
Parent: #794.

Depends on #802 — must land first.

What: engine/shadow_report.py reads heartbeat records for a shadow agent and the live agent over a window and prints: jobs, approved-versus-rejected agreement, malformed-output rate, median and p90 time per job. Fixture tests.

Accept: tests/test_shadow_report.py passes; the report runs against the live spool.

Risk: standard

Issue #803, opened 2026-09-14
Shadow plist for the review engine


nateprich:
Parent: #794.

Depends on #802 — must land first.

What: Add launchd/com.nateprich.command-center-muse-review-shadow.plist invoking scripts/muse-review-engine --shadow standard on the same cadence as the standard review job. Header documents that it runs beside the old routine and applies nothing. Extend tests/test_launchd_drift.py to know the file.

Accept: The plist file is on main and the drift test lists it.

Risk: standard

PR #864, opened 2026-09-14
scripts/muse-review-engine and the judgement-only review prompt, with --shadow (#802)

### LINKED ISSUE #802: scripts/muse-review-engine and the judgement-only review prompt, with --shadow
Parent: #794.

Depends on #799, #800, #801 — must land first.

What: New runner beside scripts/muse-review: begin, review-packet, and if precheck fails apply rejected with no model call; else run muse exec with --disable-shell --disable-write --disable-web-tools, the packet inline, and routines/muse-review.md (judgement text only, under 500 words: does the diff do what the ticket and plan say and avoid what the plan rejected). Parse, retry once on parse error, then review-apply. `--shadow` records the answer on the heartbeat finish note and as an agent comment on the PR instead of applying. Runner tests with a fake muse binary.

Accept: tests/test_muse_review_engine.py passes; routines/muse-review.md is under 500 words; a shadow run applies nothing and records its answer.

Risk: escalated

nateprich:
Closes #802 (parent #794, Phase 1).

What this does: new runner scripts/muse-review-engine beside scripts/muse-review. begin claims the PR, review-packet assembles the evidence, and a failing precheck is applied as rejected with no model call. Otherwise the model sees routines/muse-review.md with the packet inline under --disable-shell --disable-write --disable-web-tools, its one JSON answer is validated with one retry on parse error, and review-apply performs every side effect. --shadow validates (--validate-only) and records the answer as an agent comment on the PR plus the heartbeat finish note, applying nothing.

Continued the existing remote ticket/802 branch (no PR was open, so no verdict to address); did not reset. Established contents before continuing: the 532-line runner, the 55-line prompt, the --validate-only apply path, and both test files, all scoped to #802. Rebasing onto current main only restored the #815/#816 implement files the stranded branch predated; the diff vs main is now exactly the six #802 files.

The stranded work was committed as unverified and had a real bug: both the live and shadow paths read $? after an if whose condition had failed, which is always 0 in bash, so the retryable exit 3 never reached the retry branch and every malformed first answer fell through to final handling. Fixed by capturing apply_status and validate_status from bare calls (commit b842d4e4), plus an exist_ok fix in the test harness. No test expectation was weakened.

Accept evidence:
- tests/test_muse_review_engine.py + tests/test_engine_review_apply.py: 133 passed.
- Full suite: 1653 passed, 1 skipped. The only errors in either run were the heartbeat_isolation teardown guard firing because a live scheduler appended to the real spool mid-run (tracked by open #601); every errored test passes in isolation.
- routines/muse-review.md is 383 words total (313 after the ---), under 500.
- bash -n clean on the runner; funnel.py compiles.

Deliberately not done: no shadow plist (that's #803), no shadow-report (#805), no cutover (#806). routines/muse.md untouched -- the old routine runs until cutover. muse-review.md not added to test_routine_contents ROUTINES: it is judgement text with no finish/run-id language by design.

One thing to flag, not a departure: --shadow needs a way to validate without recording, so review-apply grew a --validate-only flag (decision JSON on stdout, same 3/1 exit split, records nothing). The ticket names the shadow behavior; this is the mechanism, pinned by four new apply tests.

Unsure: nothing structural. The shadow comment shape (marker + decision + capped answer) is new surface #805 will read; if the agreement read wants different fields, that is a #805 tweak.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "b842d4e4339576b15f1bd1a0089c05ebce2e7d7e",
  "reviewed_at": "2026-09-14T18:22:06.731258+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-14T18:22:08.563078+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #801, opened 2026-09-14
Merge-safety read replaces the brief in the merge path


nateprich:
Parent: #794.

What: engine/merge_safety.py computes every merge prerequisite fresh and fail-closed: ticket-to-project binding and Building status, CI green, verdict at head, rejected-merge counter, conflicting branch. `funnel merge` reads it instead of calling brief. Nothing else is computed. Fixture tests; a timing test that it completes without loading any reporting section.

Accept: `funnel merge` no longer invokes brief (grep and test); merge refuses on each prerequisite in fixtures.

Risk: escalated

Issue #643, opened 2026-09-12
Funnel dashboard on nateprich.com (Cloudflare), read-only against GitHub


nateprich:
# Funnel dashboard on nateprich.com, read-only against GitHub

Class: New (set by Nate 2026-09-12).

## What it is

A private web page at `funnel.nateprich.com` that Nate can open from any device. The top half renders the `/funnel` brief: what is waiting on him, in `funnel.py`'s order, plus the brief's diagnostic sections. Below that is a board of every parent project grouped by stage. The page never writes to GitHub, never ranks anything, and never opens a path into the Mac mini.

**Data flows outward only.**

1. **Spool.** Every successful `funnel brief`, including the ones agents already run before merging, writes its output to a local spool on the Mac, alongside a board payload computed from the Project items that run already loaded. Nothing new is read from GitHub.
2. **Publisher.** A launchd publisher job on the Mac holds the only Cloudflare credential and pushes the newest spooled snapshot to Cloudflare storage.
3. **Page.** A Cloudflare Worker behind Cloudflare Access serves the page from that snapshot and shows the snapshot's age.
4. **Refresh.** A refresh button on the page sets a flag in Cloudflare storage. The publisher polls for it (a Cloudflare call, not a GitHub call) and runs one `funnel brief` only when he has asked.

**Board v1:** parent projects only, grouped Ideas → Shaped → Ready → Building → Parked, plus Done closed in the last 7 days. Columns are short repo name (no owner prefix), title linking to the issue, Class, time at stage, and tickets closed/total. Order within a stage comes from `funnel.py`.

**Revision:** ship a thin v1, then revise it live. Each round of Nate's feedback becomes new tickets under #643, which the funnel reports as "ticket added after Building". #643 closes only when Nate accepts it.

## The work

In order. Breakdown sizes the tickets.

1. **Spool writer in `funnel.py`.** After a brief succeeds, write `{brief, board, generated_at}` to a local spool file. `board` is built from the already-loaded items, with ordering computed in Python. A spool write failure must never change the brief's exit code or output. A failed or partial brief (non-zero exit) is not spooled, so the page keeps the last good snapshot and marks the failure. Brief stdout and its `/funnel` contract are unchanged. Tests cover each of these behaviours.
[...]

PR #915, opened 2026-09-16
GitHub webhooks refresh the dashboard; scheduled brief as backstop (#914)

### LINKED ISSUE #914: Nothing produces dashboard snapshots on a schedule, so the page can be hours stale
Parent: #643.

What: the publisher uploads whatever the spool holds and only runs a brief itself when the page's refresh flag is set and the newest entry is older than ten minutes. The spool is filled by whichever agent run happens to call `funnel brief` — and #801 and #824 deliberately took the brief off the agents' merge path, so almost nothing writes it now. Measured 2026-09-15: the live page served a 19:21 snapshot at 20:30, and the newest spool entry was 79 minutes old. Nate: GitHub Projects updates in near real time; this should not take an hour.

Change: the publisher runs one brief whenever the newest spool entry is stale, not only when a refresh was tapped, keeping the existing single-brief-per-tick bound, timeout and fail-open behaviour. A tick that is already fresh still costs nothing.

Accept: with no refresh taps and no agent runs, the published snapshot is never older than about ten minutes plus one brief; the publisher log shows a scheduled brief; fixture tests cover stale-without-flag, fresh-without-flag and the flag path.

Risk: standard

nateprich:
Closes #914. Part of #643.

**Why the page could be an hour old.** The publisher uploads whatever the spool holds and only ran a brief when the refresh flag was set *and* the snapshot was already 10 minutes old. The spool was filled by agent runs calling `funnel brief`, and #801/#824 deliberately took that off the merge path, so nothing produced snapshots on a schedule.

**Webhook.** The Worker accepts `POST /api/github-webhook`, handled before the Access check because GitHub cannot log in through Access. It verifies GitHub's `X-Hub-Signature-256` HMAC over the raw body, refuses every request when no secret is configured, answers `ping`, ignores events that change no funnel state, and otherwise sets the same refresh flag the page's button uses. It reads nothing.

**Publisher.** A refresh runs a brief once the snapshot is older than 90 seconds, so a change reaches the page in about a minute and an event storm still costs at most one brief per 90 seconds. With no flag, a brief runs after 30 minutes as the backstop.

**Cost.** One brief is 40 GraphQL points and about 57 seconds of Mac CPU, measured twice. Idle hours now cost 2 briefs (80 points, 1.6% of the hourly limit) instead of nothing-then-hours-stale; a busy hour is bounded by the 90s floor.

**Still to configure** (separate steps, not in this PR): the Worker secret, the repo webhooks, and one Cloudflare Access bypass for the webhook path, which the deploy token cannot do.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "3deb0ffb359307e6906d9b4341bbe6673ebb1822",
  "reviewed_at": "2026-09-16T04:17:34.699719+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-16T04:17:36.334358+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #902, opened 2026-09-15
Dashboard v2: GitHub-Projects-style board with PR, tier and owner flags; drop the text sections


nateprich:
Parent: #643.

What: Nate, 2026-09-15, looking at the live page: the text sections are unusable and he does not want them. Remove every brief text section from the page except human steps (no missing-sections, checkout changes, machine-local steps, projects closed by the funnel, blocks cleared, unattended approvals, and the rest). Render the board like the GitHub Projects table he uses: grouped by Status, one row per project with title, repository, class, sub-issue pips and updated age. Add three flags per row: PR status (blank, submitted, approved), tier (standard or escalated), and who owns the next step (Nate, Claude, Muse, Codex). Colour the sub-issue pips: light blue for a ticket whose PR is submitted, light purple for approved, solid for closed, muted for not started.

Server side: dashboard_board() gains per-ticket rows built from items already loaded plus the PR facts the brief already read, with a bounded verdict lookup for open ticket PRs only. No new per-ticket GitHub reads.

Accept: the live page shows the board and human steps and nothing else; the flags and pip colours match a hand-check of two projects; fixture tests cover the ticket rows and the owner rules.

Human step: a Claude Code environment

Risk: standard

PR #917, opened 2026-09-16
The page polls its snapshot and keeps expanded rows open (#902)

nateprich:
Closes #902. Part of #643.

An open tab showed whatever was current when it loaded. It now polls `/api/snapshot` every 30 seconds, re-rendering only when `generated_at` changes, so with the webhook in place a change in GitHub appears on an open tab within about two and a half minutes without touching anything.

- The poll is a KV read through the Worker: no GitHub budget, no brief.
- A hidden tab does not poll, and refreshes once when it becomes visible.
- Expanded projects are remembered by ref, so a re-render does not collapse the row you were reading.

The Refresh button stays for now: unlike a reload it asks the publisher for a *new* brief, which is the only way to pick up time-based fields when nothing in GitHub has changed. Say the word and I will drop it.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "a1602ca60235914ca244ef8423d71f16ce31d505",
  "reviewed_at": "2026-09-16T04:35:03.590277+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-16T04:35:05.592609+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #921, opened 2026-09-16
Page assets revalidate, so a deploy reaches an open browser (#902)

nateprich:
Closes #902. Part of #643.

The board did not change even on a reload. The snapshot published at 05:26Z and the Worker deployed at 05:21Z, so data and code were both current; what was stale was the browser's copy of `/app.js`. The Worker set `cache-control` on its JSON responses only and passed asset responses through untouched, so the browser reused the script it had and rendered new data with old code — which is what a 44-segment bar looks like after the proportional-bar change.

Every response through `withSecurityHeaders` now carries `no-cache, must-revalidate`. The asset layer still answers 304 from the ETag, so a reload costs a conditional request rather than a download.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "3ad37ad468dd53577421139bcb2e5934f9201b16",
  "reviewed_at": "2026-09-16T05:33:00.699993+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-16T05:33:02.691391+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #935, opened 2026-09-16
The dashboard says when a PR state could not be read (#934)

### LINKED ISSUE #934: The dashboard cannot tell 'no PR' from 'PR scan failed'
Parent: #643.

What: Nate, 2026-09-16: the PR column is empty on every row and he has never seen a submitted or approved pip. Two causes, one of them a real defect. There are genuinely no open ticket PRs at this moment, so the column is truthful right now. But the 07:15 snapshot also records `ticket_pr_facts` as degraded with "GitHub read timed out", and when that scan fails the brief hands the board an empty mapping — indistinguishable from "nothing has a PR". Every row then reads as no PR whether or not one exists, silently.

Change: (1) the brief tells the board whether the scan succeeded, and a ticket whose PR state could not be read reads `unknown` rather than null, rendered as a muted `?` with a tooltip rather than a dash; (2) raise the `ticket_pr_facts` budget from 8s to 20s, since it is two bounded reads per member repo and is now the dashboard's only source of review state; (3) fixture tests for both the unknown path and the ordinary path.

Accept: with the scan failing, the PR column reads `?` on open tickets and the snapshot still renders; with it succeeding, an open ticket PR reads submitted or approved; tests cover both.

Risk: standard

nateprich:
Closes #934. Part of #643.

Nate: the PR column is empty everywhere and he has never seen a submitted or approved pip.

**Two findings.** There are genuinely no open ticket PRs across the five member repos at this moment, so the column is truthful right now. But the 07:15 snapshot records `ticket_pr_facts` as degraded ("GitHub read timed out"), and on that path the brief hands the board an empty mapping — indistinguishable from "nothing has a PR". Every row then reads as no PR whether or not one exists, silently, which is the thing the honesty rules exist to prevent.

- The brief passes `pr_facts_known`; an open ticket whose state could not be read reads `unknown`, renders as a muted `?` with a tooltip, and takes its own bar segment rather than colouring as untouched.
- A caller that passes no facts at all keeps the old reading, so fixture-pure callers are unchanged.
- `ticket_pr_facts` budget 8s → 20s. It is two bounded reads per member repo and the only source of review state the page has; 8s was losing it routinely.

Tests: failed scan reads unknown and shows in the bar; successful empty scan still reads as no PR.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "4f168a408e022700283ea0acc5184b73456fc232",
  "reviewed_at": "2026-09-16T07:30:47.060658+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-16T07:30:48.997388+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #953, opened 2026-09-16
finish-ticket runs a Makefile's python with the runner's interpreter


nateprich:
Parent: #794.

What: FF-Weekly-Start-Sit#81 errored three Codex runs in a row (11:45, 12:05 and 12:24Z on 2026-09-16) with `tests failed: make check test failed: make: *** [check] Error 1`, while the same `make check test` passed 184 tests under Codex's Python 3.12. `run_tests` swaps `sys.executable` in only when a command's argv[0] is `python` or `python3` (#890). A resolved `make check test` keeps the Makefile's `PYTHON ?= python3`. Inside Codex's sandbox that is Apple's `/usr/bin/python3` 3.9, whose `compileall` raises `PermissionError` writing its cache (Codex transcript and automation memory, run `6d48cbe75652`). `compileall` prints its errors on stdout, and `_run` reports stderr when stderr is non-empty, so the note kept only make's last line. Fix: run test commands with `PYTHON` set to `sys.executable` in the environment, so a Makefile's `?=` default resolves to the runner's interpreter, and report stdout as well as stderr when a command fails.

Accept: a resolved `make …` test command runs with `PYTHON` set to the runner's interpreter; a failing command's error carries both its stdout and stderr; the full suite passes; FF#81's next Codex finish opens its PR.

Human step: a Claude Code environment

Risk: standard

Issue #890, opened 2026-09-15
finish-ticket push is rejected after a rebased ticket branch, and CI-resolved bare python is missing on the Mac


nateprich:
Parent: #794 (Phase 3).

What: `finish-ticket` fails on the schedule Mac for two runner-environment reasons, and each one wedges a ticket in a loop.

1. **Push rejected after a rebase.** When a ticket PR is rejected as stale ("rebase and re-offer"), the implementer continues the remote `ticket/<n>` branch, rebases it onto `origin/main` as the verdict asks, and `finish_done` then runs a plain `git push --set-upstream origin ticket/<n>`, which the remote rejects as non-fast-forward. Measured 2026-09-15: command-center#884 errored five runs in a row (08:18Z, 08:51Z, 09:58Z, 10:31Z, 11:04Z) with `! [rejected] ticket/884 -> ticket/884 (non-fast-forward)`, and #866 hit the same at 06:17Z. The live workspace `run.KeXymj` reflog shows `rebase (start): checkout main` over `efa70dc9` (the tip of the rejected PR #889). Each failure discards ~17 minutes of Muse work and leaves PR #889 stale.
2. **Bare `python` from a CI step.** A test command resolved from a member repo's CI step runs verbatim, and GitHub's `setup-python` provides `python` where the Mac has only `python3`. Measured 2026-09-15 07:38Z: jeffy-finance-agent#65 errored `[Errno 2] No such file or directory: 'python'` (jeffy's `tests.yml` step runs `python -m pytest -q scripts/tests`).

Change: (1) before either push in `engine/implement.py` (`finish_done` and `_keep_work`), fetch the remote ticket branch and, when it exists and is not an ancestor of `HEAD`, record it with `git merge -s ours --no-edit origin/ticket/<n>` so the push is a fast-forward that keeps the run's tree. No force-push; the old commits stay in the branch history, and main squash-merges. (2) When a resolved command's program is `python` or `python3`, run it with `sys.executable`.

Accept: fixture tests with a real bare remote show a rebased ticket branch pushing successfully with the run's tree on the remote tip, and a resolved `python -m pytest` command running under `sys.executable`; the next Muse run on #884 finishes `done`.

Risk: escalated
Human step: a Claude Code environment

Issue #169, opened 2026-09-08
Stranded-work detector reported in funnel brief as a diagnostic array


nateprich:
Parent: #133 (piece 2). Independent of the piece-1 ticket (#168) — workable in any order; no dependency on #161.

What to build: a derived function over open tickets — is there any agent or gate that could act on this? If not, it is stranded, and the reason is reported with it. Report in funnel brief as a diagnostic array beside stale_locks_taken_over and working_tree_touched, and NOT in total_needing_nate (stranding is a machine defect, not a decision waiting on Nate — #85).

First release covers only shapes derivable from facts already fetched:
- an approved verdict against an unmergeable branch (piece 1 should make this extinct — keep the check anyway, it is the proof),
- a claim past its TTL with no PR,
- a parent left Building with no children after #65's trap,
- an item blocked on a blocker that will never close.

Constraints from the plan:
- Derived check, no new state: no cache, no .json of record (AGENTS.md:37, GitHub is the state).
- Deferred, state it rather than half-doing it: a PR whose CI never reported needs check-run history this code does not fetch — do not guess from a missing status, which would alarm on every PR mid-run.
- Do not fix Codex recording errored for a correct refusal — that is #129's.
- Rejected alternatives that must stay rejected: no auto-rebase, no auto-close of superseded PRs, no nagging in the gate queue.

How to tell it worked (fixtures only, no network):
- funnel brief reports a stranded item with its reason, and total_needing_nate does not change.

Risk: standard

nateprich:
Closing: the work is on `main` and verified.

PR #234 delivered the stranded-work diagnostic but carried no closing-issue
link, so this stayed open and kept re-entering the queue.

Verified now: `funnel brief` returns a top-level `stranded` array.

Same missing-link failure as #156; #177 tracks it.




```json
{
  "agent": null,
  "at": "2026-09-08T20:56:13.389770+00:00",
  "run": null,
  "voice": "agent"
}
```

PR #1025, opened 2026-09-17
finish-ticket runs a pinned-Python repo's tests under the pinned interpreter (#1024)

### LINKED ISSUE #1024: finish-ticket runs a pinned-Python repo's tests under whatever python3 started it
Parent: #794.

What: finish-ticket fails The-League tickets whenever Codex starts it with plain `python3`. Evidence:
- The-League#168 errored at 08:12Z (run `64539e867eda`) and #169 at 08:25Z (run `e187fdb302df`), both with `tests failed: make test failed: Fantasy-Football requires Python 3.12; found 3.9.6`.
- In both runs, Codex's own `make test` had passed under Python 3.12.

Cause: #953 made `run_tests` set `PYTHON=sys.executable`, so a Makefile's `PYTHON ?= python3` uses the interpreter that runs finish-ticket. Codex usually runs `python3 …/funnel.py finish-ticket`, and in its sandbox that resolves to Apple's `/usr/bin/python3` 3.9.6. The-League pins `.python-version` to `3.12`, and `scripts/check-python.py` rejects any other version. #168's retry passed only because Codex re-ran finish-ticket by hand under a 3.12 venv (transcript `rollout-2026-09-17T01-00-57-…`). The #890 swap of `python`/`python3` in argv has the same flaw.

Fix: before running the test commands, read the checkout's `.python-version`, taking its first line and using major.minor only.
- If the pin matches `sys.executable`, keep today's behaviour.
- Otherwise use the first of these that exists and reports that version: `python<M.m>` on PATH, `/opt/homebrew/bin/python<M.m>`, `/usr/local/bin/python<M.m>`. Use it for both `PYTHON` and the argv swap.
- If none is found, keep `sys.executable`, so the repo's own check still names the mismatch.
- Keep the #953 guard (`PYTHONDONTWRITEBYTECODE`, no pytest cache).

Accept:
- Fixture tests cover four cases: no pin keeps `sys.executable`; a pin equal to the runner's version keeps it; a pin the runner does not match picks a matching interpreter for both `PYTHON` and argv; an unsatisfiable pin falls back to `sys.executable`.
- The full suite passes.
- The next Codex finish of a The-League ticket started with plain `python3` gets past the version check.

Risk: standard

nateprich:
Closes #1024. Parent: #794.

**Problem:** finish-ticket failed The-League#168 and #169 with `Fantasy-Football requires Python 3.12; found 3.9.6`. Since #953, `run_tests` sets `PYTHON` to the interpreter that started finish-ticket. Codex usually starts it with Apple's `python3` 3.9.6, while The-League pins 3.12.

**Change:** `pinned_interpreter(root)` reads the checkout's `.python-version` (major.minor) and picks the interpreter to use:
- It looks for a matching `python<M.m>` on PATH, then in `/opt/homebrew/bin` and `/usr/local/bin`, and checks what version each candidate reports.
- It falls back to `sys.executable` when the repo has no pin or no match exists.
- The chosen interpreter is used both for `PYTHON` and for the #890 swap of `python`/`python3` in argv.

**Tests:**
- Five new fixture tests cover: no pin, a pin already met, a match on PATH, a match only in the Homebrew dirs, and a binary that lies about its version.
- Four of the five fail on `main`. The no-pin test passes there by design, since it pins #953's behaviour.
- Full suite on the Mac: 1965 passed.
- Live check under `/usr/bin/python3` 3.9.6 against Codex's The-League#169 checkout returns `/opt/homebrew/bin/python3.12`.

Taken by the funnel watch because finish-ticket is in the engine path.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "7c8d3a750933a3389ac74092d6e0e5e47c1f9fb9",
  "reviewed_at": "2026-09-17T08:52:19.493956+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": null,
  "at": "2026-09-17T08:52:22.599256+00:00",
  "run": null,
  "voice": "agent"
}
```

nateprich:
**Shadow review (not applied)** — run `03dbd0ce38ce`, packet head `7c8d3a750933`

The model answered the review question.

Decision: **approved**.

```json
{"verdict": "approved", "blocking": [], "unsure": []}

```

PR #1146, opened 2026-09-19
Reviewer instruction: read scope from compare, not gh pr diff (#1137)

nateprich:
Part of #1043.

Implements #1137.

Summary:
In routines/muse.md review section, replaced the `gh pr diff <pr> --repo <repo>` scope read with `gh api repos/<repo>/compare/<base_ref>...<head_sha>` and added one sentence explaining that a PR's own diff is computed against its recorded base so it can overstate scope once the branch has merged its base. Full suite passes (2139 passed).

Departures:
- None.

Branch:
Created fresh from origin/main; no existing remote ticket branch.

Verified:
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -c 'from pathlib import Path; compile(Path("funnel.py").read_text(), "funnel.py", "exec")'`
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -m pytest tests/ -q`

Test command source:
CI .github/workflows/tests.yml step "Run the suite"

PR #1119, opened 2026-09-19
Audit Windows-routability of CI jobs and route eligible ones to the idle Windows runner (#1109)

nateprich:
Part of #1076.

Implements #1109.

Summary:
Audited command-center CI for Windows-routability: the pytest job is POSIX-bound (#!/bin/sh gh stub, bash-driven statusline.sh, /bin/bash install.sh, os.symlink, macOS launchd paths), the dashboard Node checks are portable but too small to relieve the hobby-linux queue alone, and watchdog already runs on GitHub-hosted runners. Per the ticket fallback, made no runs-on routing change and recorded the audit and the extra-Linux-processes conclusion in docs/windows-runner-audit-1109.md. Verified: pytest 2112 passed, dashboard tests 44 passed.

Departures:
- Did not inspect each member repo's workflow files individually: this checkout contains only command-center, and routing member-repo jobs needs a PR in each member repo; the POSIX-based criteria and the no-routing conclusion are recorded in the audit note.

Branch:
Created fresh from origin/main; no existing remote ticket branch.

Verified:
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -c 'from pathlib import Path; compile(Path("funnel.py").read_text(), "funnel.py", "exec")'`
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -m pytest tests/ -q`

Test command source:
CI .github/workflows/tests.yml step "Run the suite"

PR #1116, opened 2026-09-19
First-week funnel-watch cost review against setup estimate (#711)

nateprich:
Part of #685.

Implements #711.

Summary:
Added docs/funnel-watch-cost-711.md: first-week funnel-watch cost review from 28 measured claude heartbeat runs (quiet mean 70.7k / median 45.3k output tokens, wedge mean 142.6k, weekly total 3.42M tokens = 37.5 weekly-window points, ~3.5x the $50 setup estimate), with keep-Opus/high, drop-to-medium, and change-cadence options priced and no recommendation. Verified: 2112 pytest passed, 44 dashboard tests passed.

Departures:
- Did not read the #684 log comments (no GitHub access from this workspace); classified runs from the heartbeat finish notes, which carry the same per-run summaries, instead.
- Did not write the review onto parent #685 (no remote side effects permitted); the review is carried in docs/funnel-watch-cost-711.md via this PR, for Nate to record on the parent.
- Did not compare input tokens against the estimate (the heartbeat measures output only), so the input-token half of the estimate has no measured counterpart.

Branch:
Created fresh from origin/main; no existing remote ticket branch.

Verified:
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -c 'from pathlib import Path; compile(Path("funnel.py").read_text(), "funnel.py", "exec")'`
- `/Applications/Xcode.app/Contents/Developer/usr/bin/python3 -m pytest tests/ -q`

Test command source:
CI .github/workflows/tests.yml step "Run the suite"

Issue #976, opened 2026-09-17
Review shadow runs at max effort; live standard reviewer runs at high


nateprich:
Parent: #794.

What: The review shadow runs at a different reasoning effort from the live reviewer it is measured against, so #806's agreement rate compares different effort levels. `launchd/com.nateprich.command-center-muse-review-shadow.plist` passes only `--shadow standard`, so `scripts/muse-review-engine` falls back to its default effort, `max`. The live standard reviewer, `com.nateprich.command-center-muse-review-standard`, runs `muse-review standard high`. On 2026-09-16, `launchctl print` confirmed both argument lists as loaded. Both use the same model, `muse-spark-1.3`, so effort is the one knob that differs by accident. The other differences (no tools, packet-only evidence, deterministic prechecks) are the design under test.

Change the shadow plist's arguments to `--shadow standard high`, then boot the shadow job out and bootstrap it again from the run clone. Record the time the new arguments took effect.

Verdicts recorded before that time came from `max` effort and do not count toward #806. The clean window restarts at the switch time: #806's "48 hours or 50 shadow jobs" is counted from there, and its `shadow_report --since` value moves to that time. Comment the new start time on #806 and move its `Blocked until` date to match.

Accept: `launchctl print gui/501/com.nateprich.command-center-muse-review-shadow` shows `standard` and `high` in its arguments; the first shadow heartbeat after the switch appears; #806 has a comment naming the new window start and an updated hold date.

Human step: a Claude Code environment

Risk: standard

nateprich:
Installed by the funnel watch (scheduled 02:13Z run). PR #978 merged at 02:32Z. I copied the plist and bootstrapped it from the run clone at `e2f80a2b` at **2026-09-17T02:37:39Z**. `launchctl print gui/501/com.nateprich.command-center-muse-review-shadow` shows the arguments `--shadow standard high`, and `tests/test_launchd_drift.py` passes (42). The first shadow heartbeat after the switch is the shadow review of PR #982, before 02:55Z. #806 now carries the new window start and a `Blocked until 2026-09-19` hold.

PR #1179, opened 2026-09-21
Cut the standard lane over to the engine; shadow the escalated tier (#813)

### LINKED ISSUE #813: Breakdown and shape cutover after the shadow period
Parent: #794.

Depends on #812, #805 — must land first.

What: After at least 48 hours or 50 shadow jobs, record shadow-report on this ticket (agreement here is directional, decided by Nate 2026-09-17: ticket counts within one, the same needs_decision outcome, and the same Shaped-versus-Ready outcome, except that a shadow more cautious than live (held at Shaped, or asked a question) still agrees; see #1033). Only above 95 percent: swap the standard plist to the engine, bootout the shadow.

Accept: Report on the ticket meets the threshold; the standard lane runs the engine.

Risk: escalated

nateprich:
Closes #813.

The standard plist now runs `scripts/muse-review-engine standard max` instead of the old routine, and the shadow plist moves to `--shadow escalated max` on the escalated reviewer's fifteen-minute interval.

**Why both plists move together.** One plist serves review, breakdown and shape, and the engine runner takes whatever `begin` offers, so #813's cutover could not be taken without also moving standard-tier review. Nate's decision in session on 2026-09-20 was to ship the measured tier and start measuring the unmeasured one.

**The evidence at cutover:**

| | breakdown/shape (#813) | review (#806, standard tier) |
|---|---|---|
| agreement | 20 of 20 | 11 of 12 |
| engine-wrong | none | **0%**, no wrong approval |
| malformed | 1.3% | 0% |
| sample | 76 shadow jobs, 71.6 h | 12 compared, 31 h |

#806's gate is the engine-wrong rate, not raw agreement (Nate, 2026-09-16): every raw review disagreement in the window was the engine correctly rejecting work the old reviewer approved — PRs #1116, #1119 and #1146.

**Effort is `max`**, the engine's own default (#976). The shadow ran at `high` only so its agreement measured the engine rather than the effort level.

**The escalated tier keeps the old routine.** All 177 shadow reviews were standard tier, so escalated has no evidence of its own. It cuts over when it meets the same bars, which is what #806 now tracks.

The machine-local half is already applied: both labels are bootstrapped and `launchctl print` shows the new arguments. This PR makes the repo copies match, which is what `tests/test_launchd_drift.py` and the run-keeper readiness record compare against.


nateprich:
**Funnel watch, 2026-09-20 22:4x PDT — closing this in favour of #1185, same commit `1fa6b58`, re-opened on `ticket/813`.**

This head was `cutover/standard-engine`. A PR whose head is not `ticket/<n>` is reachable by no lane: the review lane never offers it and the merge gate's ticket-parent-Building check (#1042) can never pass. The measurable consequence is that it sat green from 2026-09-21T03:31:37Z with **zero** `command-center-review` comments, live or shadow, for three hours — the second PR found in that exact state this run, alongside #1124.

#813 is a ticket under #794, which is Building, so #1185 merges through the normal gate. Nothing else changed; the diff is identical.

— funnel watch

Issue #1182, opened 2026-09-21
Record every provider limit hit, and learn what each window holds


nateprich:
# Record every provider limit hit, and learn what each window holds

Proposed class: Improve

## What it is
A forward-only quota record plus per-window calibration for Muse limits, answering pacing instead of guessing. Ticket 1: at the refusal moment in the muse_quota_record path, write a structured GitHub record with reset stamp, seconds-to-reset, weekly-lattice boolean, raw refusal, and anchored window total, with fixture tests for weekly lattice, off-lattice five-hour shape, and malformed stamp degraded. Ticket 2: anchor consumption to the reset stamp from paired start and end usage readings in standard-rate units, so two to three hits calibrate what each window holds, with fixture tests for anchored totals and distinct-run counting. Conditional ticket 3 only if the panel question is yes: prompt for the panel percentage at hit time as a calibration point. No backfill of the 19 bare exec-failure rows. This Improve project waits at Accept it and never closes itself.

## Rejected
Backfilling the 19 generic exec failures: rewrites history to guess quota versus crash, forward records plus degraded unknown is auditable. Log-grep calibration later: the rotating log is the gap, GitHub-derived rows are the fix. New label or field for hits: label set stays at two and derived rows cannot go stale. Cache or state file for totals: GitHub is the state. Contributor cutover reassessment: out on Nates 2026-09-20 instruction, he raises it himself. Silent fallback on unreadable stamps: reads as good news, degraded must be explicit.

## Undecided
Class acceptance Improve versus Broken for gap 1, and whether hits prompt Nate for the panel percentage or calibrate from stamps and totals alone; both asked above.

## Siblings checked
[...]

PR #1186, opened 2026-09-21
Price Muse at the standard rate card, counted from the provider's weekly reset

nateprich:
Nate checked the account on 2026-09-20: the Muse sessions run on the **standard** model, and he has been on the **$50/month Power Usage plan since before 2026-09-13**. `usage.py` was pricing that work at the contributor card on the strength of a 2026-09-10 reading, so every Muse dollar this repo has reported was wrong by about 28x for this workload — contributor discounts a cache read to 2% of a fresh token, standard only to 12%, and 94% of the input tokens here are cache reads.

## What changed

- **Rates** → `$1.25 / $0.15 / $4.25` per 1M (was `$0.10 / $0.002 / $0.20`).
- **Weekly cap** → `$200`, calibrated from the wall the provider actually refused at: the window ending 2026-09-19 18:14 PDT ran to a 429 at **$214.02** of standard-rate compute (786.6M tokens, priced from the journals). The old $20 belonged to the old card and would refuse every fire at these rates.
- **Session reserve** → `$4.50`, the largest session measured in the trailing week ($4.34), replacing the $0.11 worst case measured at the old card.
- **The total is counted from the provider's weekly reset**, not a trailing seven days.

## Why the anchor ships with the re-rate

Re-rating alone would have halted the lanes. The cap describes one provider window; the reader totalled a trailing week, which straddles two. Measured on 2026-09-20 the rolling total read **$224.81 — 112% of the cap** — while the live window held $20.44. Every routine exits immediately when over pace, so that reading would have stopped all Muse work for days.

The weekly window opens on a fixed lattice, Monday 00:00 UTC (Sunday 17:00 PDT). Three observed resets sit on it: 2026-09-14 and 2026-09-21, both named in 429 refusals, and the 2026-09-28 the account panel displays as "Resets Sep 27 at 5:00 PM".

## Evidence it is right

The anchored meter now reads **10.22%** where the account panel read **10%** thirteen minutes earlier — the closest agreement any local meter in this repo has had with a vendor's own number.

## Notes

- The percentage tables in `LEARNINGS.md` are unaffected: they are shares of a window, not dollars. The 2026-09-10 entry keeps its original text with a dated correction appended — the same cycle re-priced from the journals is **$66.05**, a full window ~**$82.57**, not $3.09, and the plan is not priced at cost.
- These dollars are a **pacing index, not a bill**. The plan is flat; nothing the meter reports is money owed.
- The `rolling` flag is retained because the dashboard, brief and heartbeat read it. It selects the flat-ceiling path in `pace` rather than describing the window, and the code now says so.
- Recording each limit hit, so the windows can be calibrated from evidence rather than one refusal, is #1182.

Full suite green: 2,156 passed.


nateprich:
Superseded by #1192: the same two commits on `ticket/1190`, so the review lane can see them and the merge gate can pass them. This head was `rate/muse-standard-pricing`, which neither can. The branch is left in place.

PR #1193, opened 2026-09-21
Standard-tier Muse lanes run at high effort; escalated keeps max (#1191)

nateprich:
Closes #1191.

Nate, 2026-09-21: *"Not having different effort levels between standard and escalated doesn't make sense."* All four Muse schedules ran `max`. This restores the difference: **standard `high`, escalated `max`**, on both the review and the implement lanes.

| label | before | after |
|---|---|---|
| `…muse-review-standard` | `muse-review-engine standard max` | `standard high` |
| `…muse-implement-standard` | `muse-implement standard max` | `standard high` |
| `…muse-review` (escalated) | `escalated max` | unchanged |
| `…muse-implement` (escalated) | `escalated max` | unchanged |

**Why `high` is the measured setting, not just the cheaper one.** Every one of the 177 standard-tier shadow reviews behind the 2026-09-20 cutover ran at `high` (#976 made the shadow match the live reviewer). The cutover then set `max`, the engine's default, which no evidence covered. The standard implementer was created at `max` on 2026-09-18, and the provider's weekly window was spent the next day (#1189).

**The runner needed the change too.** `scripts/muse-implement` refused any effort but `max`, so a plist-only change would have exited 1 on every fire of the standard implementer. It now accepts `high` or `max` on the standard tier (`max` kept so a rollback is one word) and still refuses anything but `max` on escalated. `tests/test_muse_implement.py` covers the refused pairs and that `standard high` gets past the guard.

**Nothing is applied by hand.** The keeper installs a changed plist when the merge reaches the run clone. The ticket's Accept is checked with `launchctl print` after that.

Full suite on this head: 2149 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1185, opened 2026-09-21
Cut both review tiers over to the engine and retire the shadow (#813, #806)

nateprich:
## Second commit, 2026-09-21: the escalated tier cuts over too, and the shadow is retired

Closes #806. Closes #813.

Nate's decision in session, 2026-09-21 00:10 PDT: flip the escalated review tier to the engine now, as a **recorded override** of #806's shadow gate rather than a pass of it. The escalated-only window held 1 compared job after 3.5 hours; all 177 shadow reviews before 2026-09-20 were standard tier, so that tier's own window could only close on a single-digit sample. The reasoning and the evidence are on #806.

Commit `6fe0afb16` adds to this PR:

| file | change |
|---|---|
| `launchd/…muse-review.plist` | ProgramArguments run `scripts/muse-review-engine escalated max`; the header records the override and the rollback |
| `launchd/…muse-review-shadow.plist` | deleted: no shadow period remains in #794 |
| `launchd/…muse-review-standard.plist` | header only: drops the note that the escalated tier still runs the old routine |
| `scripts/install.sh` | shadow plist removed from the install list |
| `tests/test_launchd_drift.py` | every reviewer plist must point at `muse-review-engine`; the three shadow-only tests and the shadow constant go with the file |

The machine-local half is applied and verified with `launchctl print`: both review labels load the engine, and the shadow label is booted out with its installed plist removed, so the keeper's refresh-only install does not bring it back when this merges. Full suite on this head: **2148 passed**.

`scripts/muse-review` and `routines/muse.md` are deliberately untouched. They are the rollback until #1187's replay of The-League PR #237 rejects, and #807 (which deletes them) now waits on #1187 by a native edge.

---

## First commit, 2026-09-20: the standard lane

Same commit as #1179 (`1fa6b58`), re-opened on `ticket/813` so a lane can actually reach it.

**Why it was re-opened.** #1179 was pushed to `cutover/standard-engine`. A PR whose head is not `ticket/<n>` is invisible to the review lane and cannot pass the merge gate, whose ticket-parent-Building check is #1042. It sat green with zero verdict comments of any kind for three hours. #813 is a ticket under #794, which is Building, so this head merges normally.

**Why it matters that it lands, rather than staying a machine-local change.** `main` at `a3c97b14` disagrees with what is installed, on both plists:

| plist | `launchd/` on main | installed and running |
|---|---|---|
| `…muse-review-standard` | `muse-review standard high` | `muse-review-engine standard max` |
[...]

PR #1197, opened 2026-09-21
The keeper loads an installed label launchd does not have, and waits out a bootout (#1196)

### LINKED ISSUE #1196: The keeper's plist reload can leave a lane unloaded, and nothing loads it again
What: on 2026-09-21 the merge of PR #1185 changed `launchd/com.nateprich.command-center-muse-review-standard.plist`. At 00:35:05 PDT `scripts/run-keeper` ran `install_plist`: it copied the file, ran `launchctl bootout` while that label had a review in flight, and ran `launchctl bootstrap` immediately. The bootstrap failed (`run-keeper: reload of com.nateprich.command-center-muse-review-standard failed`; the same input/output error a hand reload hit on 2026-09-20 for the same reason). That run's readiness line recorded `reload_pending=com.nateprich.command-center-muse-review-standard`; the 00:40 line recorded `reload_pending=-`, because the flag describes one run's install attempt and nothing carries it forward. The standard review lane, which also does breakdown and shape, stayed unloaded from 00:35 to 01:13 PDT, when a session bootstrapped it by hand. Nothing reported it: `agent_health` saw the other three Muse labels finishing.

Two changes in `scripts/run-keeper`, no state file for either:

1. Heal, every run: for each plist that exists in both `launchd/` and `~/Library/LaunchAgents/`, if `launchctl print gui/<uid>/<label>` fails, bootstrap it and name it in `reload_pending` only if that also fails. The keeper's own label is skipped. This derives the pending reload from launchd instead of remembering it. A lane is paused by moving its installed plist out of LaunchAgents, as the retired shadow plist was; say so in the script's header.
2. Stop the race, in `install_plist`: after `bootout`, poll `launchctl print` until the label is gone (bounded, about 30 seconds), then `bootstrap`, retrying a few times a few seconds apart.

Out of scope: `bootout` still ends a run that is in flight. #794 accepted that a replacement affects the next run; whether to defer a reload while a label has a live pid is a separate call.

Accept: `tests/test_run_keeper.py` covers a bootstrap that fails once and succeeds on retry, a label found unloaded on a later run being bootstrapped, and the keeper's own label being left alone; `python3 -m pytest` passes. On the Mac, after the next merge that changes a Muse plist, `launchctl list` still shows all four Muse labels.

Risk: escalated

nateprich:
Closes #1196.

**What happened.** The merge of PR #1185 changed the standard reviewer's plist. At 00:35:05 PDT on 2026-09-21 the keeper copied it, ran `launchctl bootout` while that label had a review in flight, and ran `bootstrap` at once. The bootstrap failed, the same input/output error a hand reload hit the night before for the same reason. That run's readiness line said `reload_pending=com.nateprich.command-center-muse-review-standard`; the next said `reload_pending=-`, because the flag describes one run's attempt and nothing carried it forward. The standard lane, which also does breakdown and shape, was unloaded from 00:35 to 01:13 PDT, until a session bootstrapped it by hand. `agent_health` saw the other three Muse labels finishing and said nothing.

**Two changes in `scripts/run-keeper`, neither with a state file.**

1. **Heal, every run.** For each plist present in both `launchd/` and `~/Library/LaunchAgents/`, if `launchctl print` cannot find the label, bootstrap it; name it in `reload_pending` only if that fails too. The owed reload is derived from launchd rather than remembered. The keeper's own label is skipped, as is a plist that was never installed.
2. **Stop the race.** After `bootout`, poll until the label has left launchd (bounded at about 30 seconds), then bootstrap with up to three attempts five seconds apart.

**A convention this makes explicit:** to pause a lane on purpose, move its plist out of LaunchAgents, as the retired shadow plist was. An installed plist with no loaded label is read as a failed reload and healed. The script's header says so.

**Out of scope, stated in the ticket:** `bootout` still ends a run that is in flight. #794 accepted that a replacement affects the next run; deferring a reload while a label has a live pid is a separate call.

**Tests.** The `launchctl` stub in `tests/test_run_keeper.py` now remembers which labels are loaded, so `print` fails between a `bootout` and a successful `bootstrap` the way the real one does, and it no longer logs reads. Four new tests: a bootstrap that fails once and succeeds on retry is not a pending reload; an installed label launchd does not have is bootstrapped on a later run with no plist change; one that cannot be bootstrapped is reported; the keeper's own label and a never-installed plist are left alone.

**Why now.** PR #1193 changes two busy Muse plists, so its merge triggers this same reload twice.


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "5ffcd2b97efdd8495ec4fea6c5401aa91b5db04e",
  "reviewed_at": "2026-09-21T10:01:35.967856+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-21T10:01:35.969421+00:00",
  "run": "5b830b91328a",
  "voice": "agent"
}
```

Issue #1172, opened 2026-09-21
Quota-parked agent_health reads the hold instead of alarming


nateprich:
Parent: #1160.

What: in the agent_health computation behind funnel brief, read the same quota hold the runners check (MUSE_QUOTA_HOLD_FILE when set, otherwise ~/.claude/command-center-muse-quota-hold): one ISO-8601 stamp parsed to epoch seconds, with missing, expired, or unreadable meaning no hold. While hold_until is in the future, report the Muse line as parked until <ISO> in place of the silence alarm with no alarm for the parked span, keeping the last pre-park record time visible. After hold_until passes, measure silence from max(last record, hold_until). Keep the no-hold dead-lane alarm and its multiple-of-normal wording unchanged. Do not change runner park behavior, heartbeat schema/truncation, or the normal-gap baseline and 5x threshold.

Accept: future hold with an old last record shows the parked-until line with no silence alarm; expired hold with the same old record returns the normal alarm; missing hold reads exactly as today; unreadable or corrupt stamp is ignored and returns the normal alarm; just-expired hold measures the gap from the reset rather than the pre-park record; full suite green via python3 -m pytest.

Risk: standard

Issue #1160, opened 2026-09-20
A quota-parked lane writes no heartbeat, so agent_health alarms that Muse is dead for the length of the hold


nateprich:
# Agent health names the quota park instead of alarming on parked silence

## The thing

Since #1156 shipped, a lane parked on the provider-quota hold exits before funnel begin and before any heartbeat start, so it writes zero heartbeat records for the whole hold. That exit is correct about the provider. The side effect is that agent_health reads a deliberately parked engine as a dead one.

Measured across the hold that ran 2026-09-19 20:22 PDT to 2026-09-20 17:00 PDT, 20 h 38 min: five lanes logged 783 parked ticks and wrote zero heartbeat records, because the hold is checked before begin. The only skipped-provider-quota finishes in the window are the two lanes that saw the refusal before the hold was recorded. The brief at 2026-09-20 23:21Z read: normal gap 1m, nothing recorded for 19h58m, 782x normal against a 5x alarm threshold. The alarm scales with hold length and every tick behind it was the park working correctly. Three check-ins had to explain the line away by hand.

## What changes

1. The agent_health computation behind funnel brief reads the same hold the three runners check: MUSE_QUOTA_HOLD_FILE when set, otherwise ~/.claude/command-center-muse-quota-hold. One ISO-8601 stamp, parsed to epoch seconds; missing, expired, or unreadable means no hold, matching #1154 lane behavior.
2. When hold_until is in the future relative to now, the Muse line reports parked until <ISO> in place of the silence alarm, naming the stamp from the file. No alarm is raised for the parked span, however long it runs. The last pre-park record time stays visible alongside the parked status so the reader can see both facts.
3. The gap check resumes from the reset: after hold_until passes, silence is measured from max(last record, hold_until), so a lane that restarts late alarms on its post-reset lateness instead of on the nineteen parked hours behind it.
4. Held versus dead stays explicit in the output: no records with a future hold reads as parked, an expected wait needing no watch response; no records with no future hold keeps the existing dead-lane alarm and its multiple-of-normal wording unchanged.
5. Regression tests beside the existing agent_health or brief tests: future hold with an old last record asserts the parked line and no alarm; expired hold with the same old last record asserts the normal alarm returns; missing hold asserts current behavior unchanged; unreadable or corrupt stamp asserts the hold is ignored and the normal alarm returns; just-expired hold asserts the gap is measured from the reset rather than from the pre-park record. Full suite green via python3 -m pytest.

[...]

PR #1249, opened 2026-09-21
Quota-parked agent_health reads the hold instead of alarming (#1172)

nateprich:
Part of #1160.

Implements #1172.

Summary:
Teach the watchdog to tell a parked lane from a dead one.

agent_health gains `quota_hold_until()`, reading the same file the runners write — `MUSE_QUOTA_HOLD_FILE` when set, otherwise ~/.claude/command-center-muse-quota-hold — as one ISO-8601 stamp parsed to epoch seconds. Missing, empty, unreadable and unparseable all mean no hold, the safe direction: a hold nobody can understand must not silence a dead lane. Unlike the shell reader it never removes the file, because a diagnostic that deletes the evidence it read is worse than one that reports nothing, and an expired stamp is returned rather than dropped since the caller needs it.

`assess()` takes `hold_until`. While it is in the future the agent reports as `parked until <ISO> (provider quota)` in place of the silence alarm, keeping the last pre-park record time visible, and raises no silence condition for the parked span — including the sparse-history absolute-silence floor, which is the same gap seen through a different lens. Once the hold has passed, silence is measured from max(last record, hold_until): a lane that could not run was not quiet, and counting the park against it would alarm for hours after the thing was fixed.

Nothing else changes. The no-hold dead-lane alarm keeps its exact wording and its 5x-normal threshold, every other condition is untouched by a hold — a park explains a gap in records, not a run that died or three that errored — and `funnel.agent_health` reads the hold only for Muse, because it is Muse's file and says nothing about any other lane.

Eleven fixtures in tests/test_agent_health.py cover the future hold, the expired hold, the just-expired hold measuring from the reset, no-hold reading byte-identically to today, the absolute floor, errored runs surviving a park, the reader's stamp/expiry/corruption paths and that it leaves the file in place, the environment override, and the brief consulting the hold for Muse alone.

Departures:
- None.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "164459b932d8dcf6a1c7f2249c14efb0d657f802",
  "note": "Reviewed in session at Nate's direction while running every Broken item to completion. Against the ticket: the hold is read from MUSE_QUOTA_HOLD_FILE or the documented default, one ISO-8601 stamp, with missing, expired, unreadable and unparseable all meaning no hold; a future hold replaces the silence alarm with a parked-until line that keeps the last pre-park record visible; an expired hold measures the gap from the reset rather than the pre-park record; and the no-hold alarm keeps its exact wording and 5x threshold. Two judgements worth naming: the reader never deletes the file, unlike the shell reader, because a watchdog that destroys the evidence it read is worse than one that reports nothing; and a park silences only the silence conditions, so errored and dying runs still raise. Eleven fixtures, suite green, no frozen ground.",
  "reviewed_at": "2026-09-21T23:42:07.155546+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "claude",
  "at": "2026-09-21T23:42:07.753936+00:00",
  "run": null,
  "voice": "agent"
}
```

nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "164459b932d8dcf6a1c7f2249c14efb0d657f802",
  "reviewed_at": "2026-09-21T23:44:12.145444+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-21T23:44:12.146279+00:00",
  "run": "2fbfd9d7244e",
  "voice": "agent"
}
```

Issue #1207, opened 2026-09-21
Refuse non-terminal status writes on CLOSED issues


nateprich:
In funnel.py central status-write helper used by approve, shaped, and claim: read live item.state at write time (no cache, journal, lock, or state file); refuse non-terminal writes on non-OPEN issues with refusal in command output plus an issue comment; allow CLOSED to Done or Parked only (closed = state CLOSED regardless of state_reason). Do not touch gate_question, self-approvable, lane filters, or brief. Proof: fixture tests for refused approve on CLOSED-at-Shaped with output plus comment, allowed CLOSED-to-Done, allowed CLOSED-to-Parked, OPEN behavior unchanged; python3 -m pytest touched suite green.

Risk: standard

Issue #1206, opened 2026-09-21
Retire the interim TickTick task once the wake path is live


nateprich:
What: the plan's closing line -- when this ships, the interim TickTick task that opens on 4 January 2027 is deleted, because the wake date now carries that reminder. Do this only after the three code tickets have merged and after re-parking the FF#93 winner-study item with a wake date of 2027-01-04, so the reminder is never in neither place.

This needs the TickTick connector, which is only reachable from a Claude Code environment on Nate's Mac. The task is one Nate created; the authorisation to remove it is this plan, which is his, and the PR or comment that closes this ticket should quote that line rather than relying on an agent's judgement that the task is redundant.

Accept: the FF#93 winner-study item is parked with a wake date of 2027-01-04 and appears in the pending-wakes section with that date; the interim TickTick task is gone, with the plan line quoted as the authorisation; both facts recorded here.

Human step: a Claude Code environment

Risk: standard

nateprich:
Funnel watch, 2026-09-23 22:55 PDT.

**The premise moved.** All three code tickets have merged (#1203, #1204 and #1205 are closed). But FF#93 was accepted and closed on 2026-09-23, so there is no winner-study item left to re-park.

**Done (reversible half):** the rerun now has its own home, nateprich-projects/FF-Weekly-Start-Sit#279. It is captured nate-relayed with no Class, parked with `**Parked wake:** date=2027-01-04 status=Ideas`, and quotes the TickTick task's title. The reminder now sits in both places, never in neither.

**Not done:** deleting the TickTick task "Run the FF#93 winner-study rerun with Fable (after championships, before the March auction)". The plan's line "When this ships, the interim TickTick task opening 4 January 2027 is deleted" authorises the deletion. That line was written against FF#93, though, and this ticket asks that the deletion not rest on an agent's judgement. An unattended watch also does not delete from Nate's task manager.

**Blocked:** Nate to confirm FF#279 as the substitute home, or to say where the rerun belongs instead. After that, any interactive session deletes the task and closes this ticket.

nateprich:
[...]

Issue #1166, opened 2026-09-20
A gate advances a closed issue, and the result is invisible to every list


nateprich:
# Gate refuses closed issues, and the brief names status-vs-state mismatches
Proposed class: Broken

## What it is
A closed issue at Ready with zero children is invisible to needing_nate, awaiting_breakdown, stranded, in_motion, and blocked, because awaiting_breakdown filters on OPEN and nothing counts Ready with zero tickets.
Observed 2026-09-20: #908 and #909 closed COMPLETED 2026-09-16 were advanced to Ready by funnel approve --yes under the #1158 pattern; both were corrected by hand to Done, so this plan is the hole, not those two rows.
Cause: gate_question and shaped_self_approvable read plan body, origin, Class, and escalation scan, never item.state; funnel approve does not refuse CLOSED.
Fix in funnel.py with fixture tests, two halves: a central status-write guard that refuses non-terminal writes on non-OPEN issues with a visible record, and one brief mismatch section derived each run from load_items() with no extra GitHub read.
Work: 1) Add the OPEN check to the central Status writer used by approve, shaped, and claim; CLOSED may only move to Done or Parked. 2) Widen gate_question and shaped_self_approvable to read item.state so CLOSED never reports self-approvable. 3) Add the brief section listing closed-not-Done-or-Parked and open-but-Done. 4) Fixture tests for refused approve on CLOSED, allowed Done and Parked writes on CLOSED, both mismatch directions listed, OPEN behavior unchanged; full touched suite green via python3 -m pytest.
Accept: closed-at-Shaped approve refuses with output plus issue comment; closed-at-Ready appears in the mismatch section, never as breakdown work; open-Done appears too; no new state file, cache, or lock.

## Rejected
- Silent refuse with no record: recreates the invisibility this fixes and violates the hold-state rejection that every withholding carry its reason.
- Brief-only listing with no guard: leaves the next closed plan advancable by the same approve path that moved #908 and #909.
- Guard-only with no brief section: leaves any existing mismatch invisible until a hand count finds it, as this one was.
- Per-caller duplicated state checks: drifts on the next edit while one helper keeps approve, shaped, and claim identical.
- Widening awaiting_breakdown to include CLOSED: lane lists promise startable work and closed items are never startable, so this would offer dead work.
- Done-only terminal set: strands legitimately parked work whose legal closed target is Parked.
[...]

PR #1250, opened 2026-09-21
Refuse non-terminal status writes on CLOSED issues (#1207)

nateprich:
Part of #1166.

Implements #1207.

Summary:
One predicate, `status_write_refusal`, decides whether a Status write may happen, and approve, shaped and claim all ask it.

A closed issue may be recorded as `Done` or `Parked` — those are what being closed means — and as nothing else. Every other stage describes work in progress, and a closed issue has none: #1206 put a CLOSED project at `Ready`, which placed it in the startable queue and the self-approval path at once with nothing able to close it again.

The state is read live at write time via `live_issue_state`, not taken from the loaded Item, because a FunnelSession reuses its Project objects across commands and a close that happened in between is exactly the case being guarded. No cache, no journal, no lock, no state file.

When the live read cannot be made, the loaded state decides rather than the write being refused. That is deliberate and worth stating: refusing on an unreadable read would let one GitHub hiccup block every approve and every claim, while the loaded state was already CLOSED in the case this guards. The live read is the improvement, not a new dependency the funnel stops for.

A refusal is both returned to the caller — printed, exit 1 — and left on the issue as a comment, so the record of what was refused survives the session that refused it. The comment is best effort: the refusal has already done its job by not writing.

The guard sits inside `_write_status`, which engine/shape.py already uses, and is called directly by the gate and by the claim's Building write, which keep their own confirmation and local-state handling. Twelve fixtures in tests/test_closed_status_guard.py: refused approve on a CLOSED project with the message and the comment, allowed CLOSED-to-Done, allowed CLOSED-to-Parked, unchanged OPEN behaviour, a live close beating a loaded open, an unreadable read falling back to the loaded state in both directions, terminal writes never consulting the read at all, and a claim declining to drag a closed project to Building.

Untouched, as the ticket requires: gate_question, the self-approval predicate, the lane filters and the brief.

Departures:
- None.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "d17088db3f84d5eeb912eecb599e1ca4f633ef92",
  "reviewed_at": "2026-09-21T23:59:14.985801+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-21T23:59:14.986967+00:00",
  "run": "e9c46766b44c",
  "voice": "agent"
}
```

Issue #1208, opened 2026-09-21
Gate eligibility reads state so CLOSED is never self-approvable


nateprich:
Widen gate_question and shaped_self_approvable in place in funnel.py to read item.state so CLOSED never reports self-approvable; no parallel eligibility predicate, no lane-filter change. Include fixture tests that CLOSED blocks self-approval and OPEN behavior is unchanged. Proof: new fixture tests pass and python3 -m pytest touched suite green.

Risk: standard

PR #1253, opened 2026-09-22
Gate eligibility reads state so CLOSED is never self-approvable (#1208)

nateprich:
Part of #1166.

Implements #1208.

Summary:
Widen the self-approval rule in place so a closed issue can never be self-approvable, and say so when it holds one.

`self_approval_eligible` gains a `state` term. A non-OPEN state fails the predicate outright; `None` means the caller has no state to offer and leaves the rule exactly as it was, so every existing caller reads identically. It is widened here rather than checked beside the predicate for the same reason the rest of the rule lives in one place — a parallel eligibility test is a second gate that can drift from the first.

`engine/shape.decide` takes the same `state` and passes it through, and `preview_decision` supplies the item's own state, so the packet path and the shaping path stay identical. When the state holds a plan, the reason names it: `the issue is CLOSED on GitHub`, beside the existing reasons rather than in place of them.

`gate_question` already returns None for a non-OPEN item — that half of the ticket was in place before this change, and the tests below pin it so it stays that way.

No lane filter changed, and no parallel predicate added. Four fixtures: CLOSED blocks self-approval while OPEN does not, an absent state leaves the rule unchanged in both directions, decide holds a closed issue and names why, and decide on an open issue is byte-identical to decide without a state.

Departures:
- None.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "10e9c237f23f8abd35b8f22b37924bed46de015b",
  "reviewed_at": "2026-09-22T00:19:10.324162+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-22T00:19:10.325206+00:00",
  "run": "207b3e9cfee3",
  "voice": "agent"
}
```

Issue #1181, opened 2026-09-21
Exclude in-PR work from funnel queue


nateprich:
In cmd_queue, add optional pr_facts param falling back to ticket_pr_facts(items) and reusing _ACTIVE_BRIEF_CACHE when a brief is in flight (mirror cmd_next_review pattern), and pass awaiting_review(items, pr_facts) into startable(). On PR-read failure or timeout, print an explicit degraded notice following the brief degraded-section convention instead of the unfiltered list. Do not touch the begin selector path. Add regression tests: fixture ticket with an open PR is omitted from queue; unavailable PR facts produce the degraded notice. Proof: new tests pass and queue omits in-PR tickets while showing the notice on read failure. One ticket because this is a single small reporting-path change plus its tests; splitting exclusion from the notice would force the plan-rejected silent-fallback interim state.

Risk: standard

Issue #1165, opened 2026-09-20
funnel queue lists tickets whose work is already in an open PR


nateprich:
# Plan: exclude in-PR work from funnel queue, and say so when the PR read is unavailable

## What it is

`funnel queue` currently lists tickets whose work is already finished and sitting in an open PR (observed: #1145 listed under “Startable by Codex” while PR #1162 was green and open). `cmd_queue` calls `startable()` without the `awaiting_review` argument, so the printed list never excludes finished work. The `begin` lane path is unaffected (both selector call sites already pass it); this is reporting only. What it costs is trust in the number the watch, the routines, and Nate read.

## Change

1. Give `cmd_queue` the same `pr_facts` treatment `cmd_next_review` already has: an optional `pr_facts` parameter falling back to `ticket_pr_facts(items)`, reusing `_ACTIVE_BRIEF_CACHE` when a brief is in flight.
2. Pass `awaiting_review(items, pr_facts)` into `startable()` in the queue path.
3. When the PR read fails or times out, print an explicit degraded notice (the brief convention) instead of silently printing the longer unfiltered list.
4. Add a regression test: a fixture ticket holding an open PR with an assertion that `queue` omits it, plus a case asserting the degraded notice appears when PR facts are unavailable.

## Rejected

- Silently falling back to the unfiltered list on PR-read failure. Rejected because a longer list reads as good news and spends exactly the trust this fix is meant to restore.
- Blocking or emptying the whole queue when PR facts are unavailable. Rejected because the queue must stay useful on a slow read; a degraded notice preserves trust without hiding everything.
- Touching the `begin` selector path. Rejected because both call sites already pass `awaiting_review`; changing them risks lane behavior for no gain.

## Decided from precedent

- Reuse the `cmd_next_review` `pr_facts` pattern and `_ACTIVE_BRIEF_CACHE` (funnel.py:11971-11982) rather than inventing a new PR-read path.
- Follow the brief degraded-section convention (funnel.py:13287) for the unavailable-read notice.

## Decided by the agent

- Degraded notice over silent fallback, for the reasoning above; the rejected alternative is the silent unfiltered list. A degraded queue line naming the missing read is checkable by the reviewer; a silently longer list is not.
- One ticket is expected: the fix is a single reporting-path change plus its regression test. Breakdown decides finally.

## Needs you

- Exposure: nothing outstanding. Reporting-only change to printed queue output; no new credentials or reachable surface.
[...]

PR #1260, opened 2026-09-22
Exclude in-PR work from funnel queue (#1181)

nateprich:
Part of #1165.

Implements #1181.

Summary:
`cmd_queue` now excludes tickets whose work is already sitting in an open PR, and says so when it cannot.

`startable()` has always taken the exclusion as an argument; the queue listing was the one caller that never passed it, so a ticket with a finished branch and an open PR kept appearing as startable work. The read follows `cmd_next_review`'s pattern exactly: supplied by the caller when there is one, otherwise fetched once and retained in an active brief cache so a following `brief` reuses the same query.

When the PR read fails or times out, the queue prints an explicit degraded notice in the brief's convention — naming the section, the error, and what it costs — and then still prints the list. An unfiltered list with no notice was the alternative and is the wrong one: it is indistinguishable from a correct list, and the whole defect being fixed is a ticket that looks startable and is not.

The begin selector path is untouched, as the ticket requires.

Four fixtures in tests/test_queue_in_pr.py: a ticket with an open PR is omitted while its sibling is offered, an unreadable scan prints the degraded notice and still lists, a supplied scan is not refetched, and a timed-out scan degrades like a failed one.

Departures:
- None.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "226f68c8e29b1d6314764394f2595cf3f96bfc51",
  "reviewed_at": "2026-09-22T01:20:24.166293+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-22T01:20:24.167563+00:00",
  "run": "01edca47023c",
  "voice": "agent"
}
```

Issue #1220, opened 2026-09-21
Once-per-SHA retry of infra-only failed job


nateprich:
Add single retry that calls gh run rerun for the failed job once when the helper says infra and the SHA is unchanged; never on real; second red on same SHA stays red for the watch. Proof: fixtures show retry fires once per SHA per job on infra only and never on real; python3 -m pytest green.

Risk: standard

Issue #1219, opened 2026-09-21
Main-CI infra-vs-real helper plus brief section with fixtures


nateprich:
Add funnel.py main-CI helper that reads each member main head against check-runs plus gh run annotations and reports infra vs real where lanes read it. Infra is zero completed steps, or a failure annotation naming lost runner contact; all else including ambiguous or unreadable reads as real with explicit degraded wording, never silent empty. Add thin derived brief section reader naming repo, SHA, job, and infra vs real; no new label or Project field, derived each run. Proof: fixtures for zero-steps plus lost-contact reads infra, test failure reads real, unreadable annotation reads real; brief shows repo SHA job plus verdict; python3 -m pytest green.

Risk: standard

nateprich:
**Rule crossed: frozen ground edited.** PR #1246 for this ticket adds one row to `skills/funnel/SKILL.md`'s field table, documenting the `main_ci` section it introduces. `skills/` is frozen while #794 lands, and this ticket sits under #1178, not under #794 or #1044, so the freeze row applies to it.

It is unavoidable for the ticket as written. `tests/test_funnel_skill_documents_brief.py` fails on any brief key the skill does not document — the guard #147 left behind, after a brief section shipped correct and invisible for as long as anyone looked. So "add a thin derived brief section" and "do not touch `skills/`" cannot both hold.

What the edit is: one table row of documentation for a section this PR adds. It changes no instruction to any agent, no routine behaviour, and no prose #794 is replacing. Rolling it back means deleting one line.

Recorded here rather than decided quietly, because the freeze is a rule this repo means.

Worked in session at Nate's 2026-09-21 direction to run every Broken item through the funnel to completion.




```json
{
  "agent": "claude",
  "at": "2026-09-21T22:59:37.126648+00:00",
  "run": "61c2544d4d6a",
  "voice": "agent"
}
```

Issue #1178, opened 2026-09-21
A main-branch CI run that dies on a lost self-hosted runner stays red until a person re-runs it


nateprich:
# Main-branch runner-loss detector plus one safe retry

Proposed class: Broken

Risk: standard

## What it is
Run 35546576274 job 106173463211 on main at a3c97b14 failed after ten minutes with zero test output during setup-node, annotation failure naming lost runner contact; same SHA went green on manual rerun fifteen minutes later.
Main stayed red 2h12m and the merge gate reads main, so every merge waited. Nothing in review or merge reads gh run annotations, so the cause was invisible except by hand check comparing main head to check-runs.
This plan adds two halves, each shippable alone. First, a funnel.py helper main-CI check that reads each member main head against its check-runs plus gh run annotations and reports infra versus real where lanes read it.
Infra is zero completed steps, or a failure annotation naming lost runner contact. All else, including ambiguous or unreadable reads, reports as real.
The canonical helper is the single source; a short derived brief section is its thin reader naming repo, SHA, job, and infra versus real.
Second, a single retry that calls gh run rerun for the failed job once when the helper says infra and the SHA is unchanged; a second red on the same SHA stays red for the watch.
Accept: fixture for zero-steps plus lost-contact annotation reads infra; fixture for test failure reads real; fixture for unreadable annotation reads real; brief shows infra versus real with repo SHA job; retry fires once per SHA per job on infra only and never on real; full touched suite green via python3 -m pytest.
Tickets: 1) helper plus derived brief section plus fixtures. 2) once-per-SHA retry plus fixtures.

## Rejected
Capacity first (second runner, dedicated label, cancel-in-progress): #1076 owns that axis; this plan fixes diagnosis and single retry, not queue wait, so keep both.
Retry until green or retry real failures: burns runner time on real breaks and hides the signal; one retry on infra signature only, second red stays red.
Silent empty or silent drop on unreadable reads: reads as good news and repeats the blindness; unreadable reads as real with explicit degraded wording.
New label or Project field for the verdict: label set is closed at two and a stored verdict needs writes and upkeep; derived rows each run cannot go stale.
Workflow-file or runner-host fix: workflows cannot see annotations across repos and host work is #1076 provisioning; funnel.py helper keeps all readers identical.
Prose note as the mechanism: queue and merge readers cannot parse prose; the predicate is computed from run, job, step, and annotation facts.
[...]

PR #1264, opened 2026-09-22
Once-per-SHA retry of infra-only failed job (#1220)

nateprich:
Part of #1178.

Implements #1220.

Summary:
One retry of a main job that never really ran, and never a second.

`main_ci_retry(repo)` reruns the failed job when the helper from #1219 says `infra` and the SHA has not already had its retry. `main_ci_retries()` does that per member repo, and `funnel main-ci [--retry]` is the command surface: without the flag it lists each red main and its verdict, with it it performs at most one rerun per red main.

**Once per SHA needs no state file.** A rerun raises GitHub's own `run_attempt`, so a SHA that has already been retried is visible as attempt 2 and stays red for the watch. GitHub is the state and it already records this — no journal, no label, nothing to keep in step. `main_ci_row` now carries `run_id`, `job_id` and `attempt` so the retry reads the same row the brief shows.

Four refusals, each one a decision the plan makes rather than an accident:

- A `real` failure is never rerun. Rerunning a genuine break burns runner time and hides the signal, which is why the plan rejects retry-until-green.
- A second red on the same SHA stays red.
- An unreadable attempt count does not retry: without it a first failure and a second are indistinguishable, and retrying blindly is how a loop starts.
- A refused or failed rerun is reported as not-retried with the reason, never claimed as done.

Every path returns its decision and reason, so a caller can say what happened. Ten fixtures in tests/test_main_ci_retry.py plus two extended in tests/test_main_ci.py.

Departures:
- No scheduled caller is wired. The plan names the retry and its conditions but names nothing that invokes it, and I did not invent one: putting it in `begin` would add a per-repo REST read to a poll that is meant to cost almost nothing, and putting it in `brief` would make a diagnostic perform side effects, which is the shape this repo rejects. It is a command — `funnel main-ci --retry` — which the funnel watch can run and a schedule can call. Choosing that schedule is a decision, not an implementation detail.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "f0dff519a96dddfe4daee1d353759b21d8ad2289",
  "note": "Reviewed in session at Nate's direction while running every Broken item to completion. Against the plan: a real failure is never rerun, a second red on the same SHA stays red, and once-per-SHA is derived from GitHub's own run_attempt rather than a state file \u2014 which is the right mechanism here and needs nothing kept in step. Every refusal path returns its reason rather than failing silently, including an unreadable attempt count, which correctly declines rather than retrying blind. One departure is recorded plainly: no scheduled caller is wired, because the plan names none and both obvious homes are wrong \u2014 begin would add a per-repo read to a poll meant to cost nothing, and brief would make a diagnostic act. Ten fixtures, suite green.",
  "reviewed_at": "2026-09-22T02:05:08.052424+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "claude",
  "at": "2026-09-22T02:05:09.443854+00:00",
  "run": null,
  "voice": "agent"
}
```

Issue #1256, opened 2026-09-22
Carry target-repo AGENTS.md text in implement packet


nateprich:
In engine/implement.py build_packet/collect call fetch_repo_text like shape.py for the target repo AGENTS.md and include the text in the implement packet as canonical rules via thin adapter. Record missing as absent fact with no error/retry, truncate oversized with explicit marker, carry text only with no credentials, change no muse exec flags, execute no .claude/ tree, keep checkout paths runtime-root-relative with no absolute path in Git. Proof: fixtures/tests for present, missing, and oversized AGENTS.md pass plus touched-area python3 -m pytest green.

Risk: standard

Issue #1251, opened 2026-09-21
The implement engine never sees AGENTS.md: the workspace is untrusted and the packet omits it


nateprich:
# Implement packet carries target-repo AGENTS.md like shape does. What it is: extend engine/implement.py build_packet/collect to call fetch_repo_text(resolved, AGENTS.md) for the target repo, include the text in the implement packet, note absent when missing, change no muse exec flags and no .claude/ execution. Tickets: 1) packet change plus fixtures for present, missing, and oversized AGENTS.md; 2) routines/muse-implement.md one-line pointer to packet-supplied rules plus full touched-area python3 -m pytest green. Rejected: --trust-workspace flag because it would execute untrusted .claude/ trees in --disable-sandbox and reopen the collaborator exposure the idea keeps shut; failing on missing file because missing is a fact not a gap; inlining .claude/ skills because text inclusion alone fixes the symptom. Undecided: none. Siblings checked: #794 engine owner complementary packet content no dependency; #1135 and #1167 escalation routing and scan region different gates; #1161 #1164 #1165 other startable exclusions different conditions; #1158 #1163 #1166 #1169 invisibility fixes different predicates; #1125 #1126 #1182 #1183 analysis cost marker docs different axis; #1076 #1178 #1184 #1225 #1168 runner CI reserve heartbeat budgets different readers; #1148 #1149 #1180 #1202 #1233 #162 #25 wake deploy audit selector review-split pricing connector no shared mechanism; keep all as written with no dependency.

Proposed class: Broken

## Decided from precedent

- AGENTS.md is canonical for every assistant, so implement runs without it miss required rules. (source: agents_md Repository Instructions: canonical instruction file for every AI assistant)
- Repo rules in the packet must derive from GitHub per run with no cache, journal, lock, or state file. (source: agents_md GitHub is the state; statusline rate-limit cache is the only exception)
- Packet must convey the canonical test command python3 -m pytest and the secrets and vocabulary rules. (source: agents_md Verification: canonical test command is python3 -m pytest via Python)
- Packet carries rule text only and never credentials; secrets stay in env or gitignored .env via Nate. (source: agents_md Secrets: credentials from env or gitignored .env, never committed or printed)
- Agent-origin Broken plan with no open question advances Shaped to Ready unattended and closes itself at done. (source: plan_md Is the plan good? becomes conditional; Building completion rule 1 confirmed 2026-09-17)
[...]

PR #1267, opened 2026-09-22
Carry target-repo AGENTS.md text in implement packet (#1256)

nateprich:
Part of #1251.

Implements #1256.

Summary:
The implementation packet now carries the target repo's own AGENTS.md, so an implementer is shown the rules the reviewer will hold it to.

Before this, an implementer working a member-repo ticket saw the ticket and its parent plan but not the repo's canonical instruction file — the one AGENTS.md itself calls the source of truth for every agent working there. `fetch_agents_md` calls the same `engine/shape.fetch_repo_text` the shaping packet uses, so the two packets cannot drift about what a repo's rules are, and `collect` reads the **ticket's own repo** via `parent_repo`, which for a member-repo ticket is not this one.

Three facts, three fields: the text, `agents_md_missing`, and `agents_md_truncated`.

- **Missing is an absent fact, not an error.** A repo without an AGENTS.md is a repo without one; a 404 is the answer, so there is nothing to retry and the packet stays valid.
- **A read error still raises.** Missing and unreachable are different, and letting a 502 read as missing would quietly tell an implementer there are no rules.
- **Oversized is truncated with a visible marker** at 40,000 characters. A reader has to be able to tell the rules from the first part of them.

Text only, and nothing else: no credential is read, no muse exec flag changes, and nothing under `.claude/` is fetched or executed — this is one file's contents in a JSON field. `build_packet`'s new arguments default to absent, so every existing caller keeps working and reads as no text carried.

Seven fixtures in tests/test_implement_agents_md.py plus one existing collect test extended to stub and assert the new read.

Departures:
- None.

Branch:

Verified:

Test command source:


nateprich:

```json
{
  "blocking": [],
  "ci": "green",
  "head_sha": "72d8511188f1e82e7ef84a02459803b407a120b3",
  "reviewed_at": "2026-09-22T02:53:32.668630+00:00",
  "verdict": "approved"
}
```



```json
{
  "agent": "muse",
  "at": "2026-09-22T02:53:32.669846+00:00",
  "run": "47f26d11c6eb",
  "voice": "agent"
}
```

